PR #195, opened 2026-06-14
chore(benchmarks): modernize for v1.8 — feature benches, stage output, gate coverage

DemchaAV:
## Summary

The benchmark suite measured only text/table primitives — none of the v1.8
vector features (SVG import, native charts, vector paths, gradients) had a
bench, and the current-speed report blended compose/layout/render into one
number. This branch adds feature-object benches and deterministic probes over
the new surface, splits the current-speed report into per-stage timings with a
readable `summary.md`, closes a silent gap in the smoke perf gate, and removes
three redundant benches. Everything lives in the separate `benchmarks/` module
plus its docs — **no library `src/main` is touched.**

## Commits (reviewed sequentially)

1. **`chore(benchmarks): remove three redundant benchmark mains`** — drop
   `FullCvBenchmark` (superseded by the JMH `TemplateCvJmhBenchmark`),
   `GraphComposeBenchmark` (duplicated `CurrentSpeedBenchmark`'s `engine-simple`
   scenario), and `ScalabilityBenchmark` (thread sweep folded into the
   full-profile throughput run, now `1,2,4,8,16`); prune the matching
   `run-benchmarks.ps1` steps.
2. **`perf(benchmarks): persist compose/layout/render stages + a run summary`** —
   `CurrentSpeedBenchmark` writes a per-scenario stage split (`stages[]`,
   median ms) to the JSON and a `stages` CSV, plus a human-readable `summary.md`.
3. **`perf(benchmarks): diff consumes stages[] and reports added/removed
   scenarios`** — `BenchmarkDiffTool` prints a per-stage delta table and the
   scenario set-change (added/removed) between two runs.
4. **`perf(benchmarks): add SVG-import feature benches`** — `SvgJmhBenchmark`
   (path parse / whole-file icon read / icon→node) + `SvgParseAllocProbe`.
5. **`perf(benchmarks): add chart feature benches`** — `ChartJmhBenchmark`
   (bar + line + pie render) + `ChartAllocProbe` (layout-compile allocation).
6. **`perf(benchmarks): add vector-paint render-operator probe`** —
   `VectorRenderOperatorProbe` draws the same paths flat vs. gradient vs.
   translucent and counts the PDF content-stream operators each paint mode emits.
7. **`bench(jmh): add icon-ramp and mixed v1.8 showcase benches`** —
   `IconRampJmhBenchmark` (icon-placement scaling, `@Param` 8/32/128) and
   `MixedShowcaseJmhBenchmark` (one document mixing prose, inline sparklines,
   bar + pie charts, SVG icons and a gradient path) as the integration canary.
8. **`bench(gate): gate the long-token scenario and guard threshold coverage`** —
   `long-token` had no SMOKE threshold and silently escaped the gate; add one
   (10.0 ms / 256.0 MB, ~3× its observed ~3.2 ms / ~94 MB) and add
   `CurrentSpeedScenarioGateTest`, which fails the build if any scenario lacks a
   threshold. Scenario list hoisted to a static `SCENARIO_DEFS` so the names are
   testable; the six scenarios, order, descriptions and renderers are unchanged.
9. **`docs(benchmarks): finish the removed-bench cleanup and fix two stale
   Javadocs`** — sweep references the removed mains left behind (`ab-bench.ps1`
   log parsing, `performance.md`, a merged row in the module README) and correct
   two docs that overstated the code (the SVG-parse fixture has no arc command;
   `stages[]` is not carried into the median aggregate).
10. **`docs(changelog): note the v1.8 feature-object benches, stage output, and
    gate coverage`.**

## Testing

- `./mvnw -B -ntp verify -pl .` → **BUILD SUCCESS, 1380 tests** (canonical suite +
  japicmp + javadoc; the CHANGELOG/docs guards `VersionConsistencyGuardTest` +
  `CanonicalSurfaceGuardTest` stay green after the doc edits).
- `./mvnw -B -ntp -f benchmarks/pom.xml verify` → **BUILD SUCCESS, 30 tests** —
  incl. new `CurrentSpeedScenarioGateTest` and a new `BenchmarkDiffToolTest` case
  covering added/removed scenarios + stage deltas.
- New JMH benches run end-to-end: `IconRamp` shows a clean near-linear
  placement ramp (8 → 0.86, 32 → 2.74, 128 → 8.79 ms/op); `MixedShowcase`
  ~8.7 ms/op. A smoke run with the gate **enabled** passes with `long-token` now
  gated. The alloc/operator probes report deterministic counts (chart
  compile-pass ~446.8 KB; gradient paint adds `sh`+`W` per shape, alpha adds `gs`).

## Notes for review

- **Benchmark-only branch** — no library `src/main` change. The new fixtures
  (`SvgBenchmarkFixtures`, `ChartBenchmarkFixtures`) are `public` only because the
  JMH benches live in the `.jmh` subpackage; each fixture is reused by both the
  bench and its probe so the two measure identical data.
- **`stages[]` is intentionally not medianed** — `BenchmarkMedianTool` aggregates
  latency/throughput only, so a median-vs-median diff shows no stage deltas; diff a
  single-run pair for stage attribution (noted in the median-tool Javadoc).
- **No committed smoke baseline** — the verdict gate stays a local same-machine
  A/B tool (a static committed baseline would false-positive on machine variance);
  `run-benchmarks.ps1` already skips the verdict step gracefully when none exists.
  The absolute SMOKE gate (now covering all six scenarios) is the CI net.

**Lane:** test — benchmark tooling + operations docs; no canonical / shared-engine / legacy surface touched.


DemchaAV:
Closing — opened prematurely. Work continues on the chore/benchmark-modernization branch; the PR to develop is the maintainer's call once the branch is finalized.

PR #196, opened 2026-06-15
chore(benchmarks): modernize and expand the suite (v1.8 coverage, CI gates)

DemchaAV:
## Summary

A **benchmark-module-only** branch (no library `src/main` is touched) that modernizes and
expands the performance suite. It removes three redundant benches, adds feature-object
coverage for the v1.8 vector surface (SVG, charts, vector paths, gradients, images), wires
the suite's deterministic checks into the per-PR CI gate, splits the current-speed report
into per-stage timings, and adds a multi-page cross-library comparative tier with a visual
sample dump. It consolidates the earlier `benchmark-modernization` and `benchmark-coverage`
lines into one branch.

## Highlights

**Cleanup.** Removed `FullCvBenchmark` / `GraphComposeBenchmark` / `ScalabilityBenchmark`
(redundant with `TemplateCvJmhBenchmark` / the `engine-simple` scenario); the thread-scaling
tier folded into the full-profile throughput run.

**v1.8 feature coverage** (the suite previously exercised only text/table primitives) — JMH
render benches and deterministic operator/allocation probes for SVG (parse / whole-file icon /
node), charts (bar / line / pie + horizontal / stacked / donut / value-axis-min variants),
vector paint (flat / gradient / alpha / stroked / dashed), image embed+scale (`PdfImageCache`
reuse), inline sparklines, an icon-placement ramp, and a mixed-showcase canary.

**Deterministic gates in CI.** The benchmarks module's tests never ran in CI; the `perf-smoke`
job now runs them, so image-cache reuse, render-operator (F5) coalescing, the vector-paint
operator structure, and scenario/threshold coverage are build-failing gates. A `vector-rich`
scenario (charts + SVG icons + gradient) joins the gated current-speed harness.

**Richer, honest output.** The current-speed report carries a compose/layout/render `stages[]`
split + a `summary.md`; the diff shows per-stage deltas and added/removed scenarios; the median
tool carries `stages[]`. The smoke gate's GC-noisy `peakHeapMb` is now advisory (fails only on
latency). README field names corrected to match the emitted JSON.

**Comparative + realism.** `ComparativeBenchmark` gained a multi-page report tier (equivalent
content across GraphCompose / iText / JasperReports, full-width tables) plus a post-run
sample-PDF dump per library/scenario. Added a single-shot cold-start bench, a production-scale
large-table bench, and an allocation-rate / GC-pressure probe.

## Verification

- `./mvnw -B -ntp verify -pl .` → **BUILD SUCCESS, 1380 tests** (canonical suite +
  japicmp + javadoc; the CHANGELOG/docs guards stay green).
- `./mvnw -B -ntp -f benchmarks/pom.xml verify` → **BUILD SUCCESS, 39 tests** — incl. the new
  image-cache, render-operator, vector-paint, and scenario-coverage gates and the median
  stage-carry case.
- Every JMH bench runs end-to-end; a `perf-smoke` smoke run passes with the `vector-rich`
  scenario gated and `peakHeapMb` advisory.

## Notes for review

- `openHTMLtoPDF` is intentionally excluded from the comparative: openhtmltopdf 1.0.10 targets
  PDFBox 2.x and fails against the PDFBox 3.x the project uses (no PDFBox-3 release exists).
  Documented in the README.
- The JMH benches and the relative ±%-vs-baseline verdict are on-demand / local only by design
  (a per-PR JMH run is too slow; a static smoke baseline is machine-specific). The per-PR gate
  is the smoke absolute thresholds + the deterministic gate tests. `docs_per_sec` is a derived
  `1000/avg` reciprocal (documented), not measured throughput.
- The new "Run deterministic benchmark gates" CI step makes the benchmark-module test classes
  PR-blocking (all deterministic, sub-second).

**Lane:** test — benchmark module + operations docs + CI; no canonical / shared-engine / legacy surface touched.


DemchaAV:
## Review summary

High-effort review of the branch (7 finder angles → recall-biased verification). **No correctness bugs** — 10 PLAUSIBLE notes, 0 CONFIRMED. Applied four small follow-ups:

- **vector-rich scenario** — parse the SVG icon and build the gradient once (instance fields) instead of inside the per-iteration render method, so it times the render rather than a re-parse; widened its SMOKE threshold `20 → 25 ms` (charts + SVG icons vary more than the text scenarios) and documented the observed ~5–6 ms basis.
- **`ImageBenchmarkFixtures.distinctImage`** — added a seed-positioned 1px marker so each index yields byte-distinct content (the modular gradient/line colours alone could repeat at large indices), keeping the distinct-embed gate robust.
- **`BenchmarkMedianTool`** — logs a note when it omits `stages[]` because the source runs' stage scenario sets differ, instead of dropping them silently.

Confirmed **intentional, left as-is**: `peakHeapMb` advisory in the smoke gate (GC-timing noisy; the deterministic memory signal is the allocation probes); `ColdStart`'s `@Setup` (measures the cold *render* path, by design); the operator probe's whole-page counting (controlled fixture + exact `==N` assertions are self-protecting).

**Verification:** `./mvnw verify -pl .` → 1380 tests, `./mvnw -f benchmarks/pom.xml verify` → 39 tests, both BUILD SUCCESS.

PR #434, opened 2026-07-24
feat(templates): render inline links in CV entry titles and subtitles

DemchaAV:
## Why

CV project rows and body text already render inline `[label](url)` Markdown as
clickable links (since v1.6.8), but experience/education entry **titles** were emitted
as plain text — so a `[name](url)` in an entry title (a project name, an employer)
showed up literally instead of linking. Subtitles were inconsistent: link-aware in the
canonical renderer, stripped in the compact and hand-rolled ones.

## What changed

- `MarkdownInline` gains `appendTransformed(rich, text, style, UnaryOperator<String>)` —
  link-aware, applying the transform to the visible label and plain segments but never
  the URL — plus `appendUpperCased` (now a delegate) and `appendIfPresent` (the
  link-aware counterpart of `appendPlainIfPresent`).
- Entry titles and subtitles route through the link-aware helpers everywhere they
  render: the shared `EntryRenderer` and `EntryCompactRenderer` (all five variants,
  upper-case-aware), and the four presets that hand-roll their own titles —
  `MonogramSidebar`, `SidebarPortrait`, `EngineeringResume`, and `MintEditorial`
  (letter-spaced via `appendTransformed(TextOrnaments::spacedUpper)`). No `CvEntry`
  change.
- `TimelineMinimal` fuses each entry into one truncatable excerpt line and cannot carry
  a clickable link; it now strips the markdown to the label text instead of rendering
  `[..](..)` literally. The fused subtitle+date meta line in the `MintEditorial` /
  `SidebarPortrait` experience entries flattens the same way. Dates stay plain
  (out of scope).
- Backward-compatible: a title with no inline-Markdown syntax renders byte-identically —
  all CV visual-parity and layout snapshots are unchanged (no re-bless).

## Verification

`./mvnw -B -ntp clean verify -pl :graph-compose-templates,:graph-compose-qa -am` →
**BUILD SUCCESS**, qa **666 tests, 0 failures**; CV visual-parity + layout snapshots
unchanged.

- `MarkdownInlineTest` **+5**: URL preserved under upper-casing, transform hits the
  label not the URL, `appendIfPresent` emits prefix + link, blank/null is a no-op, and
  a plain title matches the old `plainText().toUpperCase()`.
- New `EntryTitleLinkTest` **(6)**: a `[label](url)` title becomes a link through
  `EntryRenderer`, the upper-cased `EntryCompactRenderer` (label upper-cased, URL
  intact), and the hand-rolled `MonogramSidebar` + `EngineeringResume` paths; the
  subtitle links too; and a plain-title control (asserted specifically, since the header
  contact block renders its own email/website links).
- A `ModernProfessional` render + PDF link-annotation check confirmed the entry-title
  URIs are present as clickable annotations in the output PDF.

**Lane:** templates (canonical). No engine or public-DSL change; links reuse the existing
`RichText.link` convention (they inherit the title's style, matching project-row links).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ follow semantic versioning; release dates are ISO 8601.
 
 ### Public API
 
+- **CV entry titles and subtitles accept inline `[text](url)` links.** In the layered
+  CV presets, a `[label](url)` in an experience/education entry title or subtitle now
+  renders as a clickable hyperlink — the same inline-Markdown convention already used
+  for project rows and body text. Upper-cased and letter-spaced preset titles keep the
+  link too, with the visible label styled and the URL preserved. Where a preset fuses a
+  line that cannot carry a clickable link — the single-line TimelineMinimal excerpt, and
+  the combined subtitle+date meta line of MintEditorial/SidebarPortrait experience
+  entries — the link's label text is shown instead. Titles without inline-Markdown
+  syntax render exactly as before, so the change is backward-compatible.
 - The fixed-layout PPTX backend ships as `@Beta` (Experimental) in its first
   release: the `document.backend.fixed.pptx` packages and the
   `DocumentSession` PPTX convenience methods (`toPptxBytes`, `writePptx`,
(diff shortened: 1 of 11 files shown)


DemchaAV (commented):
keepWithNext and link title

PR #449, opened 2026-07-26
docs: bring the release-facing surfaces up to 2.1

DemchaAV:
## Why

Three pages a visitor reaches before they ever see code, all describing an earlier release. None is rewritten by `cut-release.ps1` — the script owns version *tokens* (poms, install snippets, the hero badge, JSON-LD `softwareVersion`), and every claim below is prose it deliberately leaves alone.

## README

The stable-release banner named v2.0.0 and its module split. It now names v2.1.0 and what the release is actually about, pointing at the capability matrix instead of the 2.0 migration guide — which the README still links from six other places, so nothing is lost.

## SECURITY.md

The supported-versions table said **`1.9.x` — actively patched**. It had stayed there across two majors, while the sentence directly above it says fixes go to the latest minor release. The table said one thing and the policy said another.

## The showcase site

**PowerPoint appeared nowhere on it** — zero mentions — even though the gallery has been serving decks since the twin examples landed, and five cards already render a "Get PPTX" button from the manifest. The hero, the pipeline description, and the JSON-LD `featureList` all described a PDF-only engine.

**It told visitors to call an API they cannot obtain.** `BusinessTheme.modern()` appeared in the feature grid *and* in the JSON-LD feature list. That type ships only in the `examples` module, which is not published — and [README.md:96](README.md) has stated since 2.0 that `BusinessTheme` is gone and names its replacement. The project's two public faces contradicted each other on the theming entry point. Replaced with the real one, `BrandTheme` from `graph-compose-templates`, using factories that exist.

**Counts:**

| Claim | Reality |
|---|---|
| "14 CV presets and 14 letter pairs" | 16 CV presets, 15 with a matching cover letter (`MinimalUnderlined` has none) |
| "Cover Letter (14 paired layouts)" heading | over a `<noscript>` list of **five** entries |
| "819 green tests" | long past; the reactor is far larger |

The two `<noscript>` headers now describe the selection they head instead of claiming a catalogue total, and the test count is replaced by the gate it stood for so it cannot rot again. `50+ runnable examples` is left as-is — the manifest carries 87, so it is true and conservative.

## Deliberately not changed

The page `<title>`, `<h1>`, and `<meta name="description">` still read "Java PDF layout engine". PDF is the production backend and PPTX ships `@Beta`, so the one-line positioning is a maintainer call rather than a correctness fix — same reason the GitHub repository description and topics are untouched.

## Verification

`./mvnw -B -ntp clean verify` over the eight-module reactor — BUILD SUCCESS, 0 failures, `VersionConsistencyGuardTest` green (12 tests). Site served locally and re-read after the edits: no console errors, every one of the 39 `showcase/` links resolves on disk, no stale string survives in the rendered DOM, PPTX now appears 9 times, and the 5 "Get PPTX" gallery buttons match the 5 manifest entries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@
 </p>
 
 > **Release status** &mdash;
-> 🟢 **Latest stable**: [v2.0.0](https://github.com/DemchaAV/GraphCompose/releases/tag/v2.0.0) &mdash; the **module-first** release: the engine is now a lean `graph-compose-core` with opt-in `render-pdf` / `render-docx` backends, while `graph-compose` stays a drop-in for PDF. **[Migrating to 2.0 &darr;](docs/migration/v2.0.0-modules.md)**
+> 🟢 **Latest stable**: [v2.1.0](https://github.com/DemchaAV/GraphCompose/releases/tag/v2.1.0) &mdash; the **PowerPoint** release: `graph-compose-render-pptx` turns the same resolved layout into an editable deck &mdash; one page per slide, geometry-identical to the PDF, text and panels as native shapes. Ships as `@Beta`. **[What each backend supports &darr;](docs/architecture/backend-capability-matrix.md)**
 
 > &nbsp;·&nbsp; ⬆️ **Upgrading from 1.x?** `graph-compose` stays a drop-in for PDF with no code change; see the [2.0 modules migration guide](./docs/migration/v2.0.0-modules.md)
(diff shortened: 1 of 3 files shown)


DemchaAV:
Superseded by #450, which was branched from this branch and carried all three files with it. `origin/develop` and this branch are identical for `README.md`, `SECURITY.md` and `web/index.html`, so there is nothing left to merge here.

PR #456, opened 2026-07-26
docs(readme): show the PPTX backend, fix the architecture diagram, drop the dead star chart

DemchaAV:
## The badge row

It credited **PDFBox 3.0** and nothing for the office formats. The release is about PowerPoint output, and the library that produces it appeared nowhere in the stack row. **Apache POI 5.5** now sits beside PDFBox — it backs both PPTX and DOCX, and it is pinned at `5.5.1` identically in `render-pptx/pom.xml:60` and `render-docx/pom.xml:60`.

The CI / Release / Maven Central badges are shields.io queries, so they move to 2.1.0 on their own once the tag publishes; only Java, PDFBox, POI and License are literals.

## The architecture diagram was wrong in two directions

```
E{ServiceLoader} -->|render-pdf|  PdfFixedLayoutBackend
E{ServiceLoader} -->|render-docx| DocxSemanticBackend      ← wrong
                                  PptxFixedLayoutBackend    ← absent
```

**DOCX is not discovered through the `ServiceLoader`.** `render-docx` publishes no service file at all — the only `META-INF/services` entries in the tree are `render-pdf` (`FixedLayoutBackendProvider` + `FontMetricsProvider`) and `render-pptx` (`FixedLayoutBackendProvider`). The semantic exporter is named directly by the caller: `session.export(new DocxSemanticBackend(), target)`. `scripts/release-smoke/README.md` already states this correctly; the diagram did not.

**PPTX was missing**, though it is the backend the seam actually discovers alongside PDF.

DOCX now branches off the **node tree** rather than the engine, because that is what it consumes: `SemanticBackend.export` takes a `DocumentGraph`, whose own Javadoc calls it the *"Semantic authoring tree for the v2 pipeline"*. It never sees a layout pass — which is the reason it loses geometry, now visible in the shape of the diagram instead of only in prose.

The sentence above the diagram claimed *"each render backend is a separate module that registers through a `ServiceLoader` seam"* — true for the fixed-layout pair, false for DOCX. It now separates the two paths and says why a deck matches the PDF geometrically.

## Verification

`CanonicalSurfaceGuardTest`, `DocumentationCoverageTest`, `VersionConsistencyGuardTest`, `PackageMapGuardTest` — 27 tests, 0 failures. Both new shields URLs return `200 image/svg+xml`. The mermaid change adds one node and one dotted edge using constructs already present in the diagram GitHub renders today.

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,13 +350,3 @@ See [CONTRIBUTING](./CONTRIBUTING.md) for the branch-routing table and the full 
 ## License
 
 MIT &mdash; see [`LICENSE`](./LICENSE).
-
-## Star History
-
-<a href="https://www.star-history.com/?repos=DemchaAV%2FGraphCompose&type=date&legend=top-left">
- <picture>
-   <source media="(prefers-color-scheme: dark)" srcset="https://api.star-history.com/chart?repos=DemchaAV/GraphCompose&type=date&theme=dark&legend=top-left" />
-   <source media="(prefers-color-scheme: light)" srcset="https://api.star-history.com/chart?repos=DemchaAV/GraphCompose&type=date&legend=top-left" />
-   <img alt="Star History Chart" src="https://api.star-history.com/chart?repos=DemchaAV/GraphCompose&type=date&legend=top-left" />
- </picture>
-</a>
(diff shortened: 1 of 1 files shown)


DemchaAV:
Two more README corrections landed on this branch.

**The star-history chart is removed.** It had been rendering as broken alt text. Not a wrong URL — `api.star-history.com` answers **500** for this repo *and for its own*, with `Repo demchaav/graphcompose: timeout of 10000ms exceeded` from its GitHub call. Three retries at a 45s timeout, all 500.

The usual alternative is worse: `starchart.cc` answers **200**, but the SVG body is

```
<text fill="red">failed to talk with github api: {"message":"Requires authentication", … "status":"401"}</text>
```

— it would paint a red error banner where the chart used to be. A section that can only render broken is worse than no section, so it is gone rather than swapped.

**`docs/architecture/package-map.md` was audited and is correct** — no change needed. Every one of the 81 production packages across the six published modules is covered (the four preset families are excluded by the file's own note deferring them to the template guides), all 33 packages it names still exist, and the subtle core/render-pdf split it describes is real: `document.backend.fixed.pdf.options` does ship from `core` while the impl ships from `render-pdf`.

Worth noting for anyone reading the file on GitHub: `main`'s copy **is** stale — it still lists `engine.core` and `engine.pagination`, both deleted, and lumps DOCX and PPTX together as "POI semantic exporters". `develop` already fixed all of that; it reaches `main` with the release.

PR #464, opened 2026-07-26
fix(engine): let the keepWithNext lookahead see keepTogether

DemchaAV:
## Why

Two pagination rules disagreed, and the disagreement stranded exactly the heading the pair exists to protect.

`keepWithNext()` decides by asking whether the heading plus the **first line** of the next block fits the remaining page ([`LayoutCompiler:330`](https://github.com/DemchaAV/GraphCompose/blob/develop/core/src/main/java/com/demcha/compose/document/layout/LayoutCompiler.java#L330)). A `keepTogether()` block has no first line to break after — it relocates entire ([`:261`](https://github.com/DemchaAV/GraphCompose/blob/develop/core/src/main/java/com/demcha/compose/document/layout/LayoutCompiler.java#L261)). `leadingUnitHeight` measured the first line anyway, the lookahead concluded it fit and left the heading in place, and `compileNode` then moved the whole block to the next page.

`leadingUnitHeight`'s own Javadoc already promised the right behaviour — *"the whole outer height for any other leaf or an atomic (row / stack) composite that cannot split"*. But its atomic branch tests `layoutSpec.axis() != VERTICAL || children.isEmpty()`, so a section made atomic **by request** never reached it.

Sequence at a 150pt filler on a 360pt inner page: lookahead counts 30 (heading) + 6 (spacing) + 100 (body's first shape) = 136 ≤ 210 remaining, so the heading stays; 174pt remain; the body's `outerHeight` is 200 ≤ 360 so `keepWhole` holds, 200 > 174, `newPage()`. Body on page 1, heading on page 0.

## Who it hit

Every heading above a block that asks to be kept whole. In `graph-compose-markdown` that is five unconditional `keepTogether()` calls in `BuiltinRenderers` — code blocks, quotes, alerts, callouts, and the unsupported-block panel — and it is the one real orphan in the viewer samples (`kitchen-sink.md`, page 2, `## Blockquote`). In this repo it is the CV and proposal presets, where a section heading precedes a kept-whole card.

## The fix

`leadingUnitHeight` returns the whole outer height when the node's keep-together request will actually be honoured. The condition mirrors `keepWhole`, so a block **taller than the page** — whose request `compileNode` ignores — keeps being estimated by its first slice. That is what lets a heading above a page-spanning paragraph or table still relocate, and `SectionKeepWithNextTest` already guards it.

`leadingUnitHeight` had no view of `CompilerState`, so it takes the page's inner height as a parameter instead. A measurement helper stays free of the mutable page cursor, and the value threads through the one recursive call so a nested kept-whole child is caught too.

No new API: `DocumentNode.keepTogether()` has existed as a default method since 1.8.0, and the changed method is private with two call sites.

## Verification

**Reproduced before fixing.** At 300×400 with 20pt margins and a 200pt two-part body, the heading stranded for **every filler height from 120 to 210pt**, and never once when the same body could split.

**The test sweeps that range rather than sampling it** — a single sample passes against the unfixed compiler for most filler heights, which is how this survived. Proven to drive the behaviour: on the unfixed engine it fails at `filler=120` with the message naming it; on the fixed engine 4/4 pass. The other three cases hold in **both** directions, so they guard what this change must not alter — the splittable body, the page-spanning body whose keep-together is inert, and the best-effort boundary where heading and block cannot share even an empty page.

| gate | result |
|---|---|
| new test, unfixed engine | 1 failure at `filler=120` |
| new test, fixed engine | 4 / 4 |
| full eight-module reactor `clean verify` | BUILD SUCCESS |
| `javadoc:javadoc -pl :graph-compose-core` | BUILD SUCCESS |
| qa snapshots + visual regression | 675 / 675, zero baseline movement |

Zero baseline movement is the point: no committed layout changes, because no existing baseline exercised this path. `SectionKeepWithNextTest` mentions `keepTogether()` only in prose, and only for bodies taller than a page where the flag is inert — which is why the suite stayed green over the bug.

## Residual limitation

When the heading and the whole block cannot share even an empty page, the rule declines to relocate rather than strand the heading alone. That matches `keepWithNext`'s documented best-effort semantics and is now pinned by a test instead of left to be rediscovered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,21 @@
 All notable changes to GraphCompose are documented here. Versions
 follow semantic versioning; release dates are ISO 8601.
 
+## v2.1.1 — Planned
+
+### Fixed
+
+- **A heading no longer strands above a block that was asked to stay whole.**
+  `keepWithNext()` decides by asking whether the heading plus the *first line* of
+  the next block fits, but a `keepTogether()` block has no first line to break
+  after — it relocates entire. The lookahead measured the first line anyway,
+  concluded it fit, left the heading in place, and the compiler then moved the
+  whole block to the next page. It affected every heading above a kept-whole
+  block: quotes, code blocks, callouts and alerts in the Markdown renderer, and
+  the card sections of the CV and proposal presets. A block taller than a page
+  still anchors by its first line, because its keep-together request is ignored
+  anyway.
+
 ## v2.1.0 — 2026-07-26
 
 ### Highlights
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,11 +12,11 @@ follow semantic versioning; release dates are ISO 8601.
   the next block fits, but a `keepTogether()` block has no first line to break
   after — it relocates entire. The lookahead measured the first line anyway,
   concluded it fit, left the heading in place, and the compiler then moved the
-  whole block to the next page. It affected every heading above a kept-whole
-  block: quotes, code blocks, callouts and alerts in the Markdown renderer, and
-  the card sections of the CV and proposal presets. A block taller than a page
-  still anchors by its first line, because its keep-together request is ignored
-  anyway.
+  whole block to the next page. It affected any heading above a block that opts
+  into `keepTogether()` — including a timeline built with
+  `keepEntriesTogether()`, whose entries are kept whole one level down. A block
+  taller than a page still anchors by its first line, because its keep-together
+  request is ignored anyway.
 
 ## v2.1.0 — 2026-07-26
 
--- a/core/src/main/java/com/demcha/compose/document/layout/LayoutCompiler.java
+++ b/core/src/main/java/com/demcha/compose/document/layout/LayoutCompiler.java
@@ -327,7 +327,8 @@ public final class LayoutCompiler {
                                   + toMargin(children.get(k).margin()).vertical()
                                   + layoutSpec.spacing();
                     }
-                    needed += leadingUnitHeight(children.get(runEnd), childRegionWidth, prepareContext);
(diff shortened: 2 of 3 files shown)


DemchaAV:
Self-review found one of the four tests was a tautology; fixed in `7cd7b89c`.

`pageSpanningKeepTogetherBodyStillAnchorsByItsFirstSlice` pinned a **300pt** filler. Probing that document with and without the opt-in:

```
filler=300 kwn=false  header=p0 body=p0 glued     <- passes either way
filler=310 kwn=false  header=p0 body=p1 ORPHAN
filler=310 kwn=true   header=p1 body=p1 glued     <- where the rule fires
```

At 300 the body's first line still fits under the header, so the two land together whether or not anything relocates them. The assertion held against an engine with no `keepWithNext` at all — it proved nothing about the branch it names. The live window is 310–320 and I missed it by one step.

It now builds at 310 and asserts both directions, matching `SectionKeepWithNextTest`'s shape.

**Mutation-checked on both sides of the condition**, which is the point of having these two tests at all:

| mutation | red test |
|---|---|
| remove the whole `keepTogether` branch | `headerStaysWithAKeepTogetherBodyAtEveryFillerHeight` @ `filler=120` |
| remove the `<= pageInnerHeight` guard | `pageSpanningKeepTogetherBodyStillAnchorsByItsFirstSlice` |

The second mutation passed the 300pt version, so that guard was unprotected until this commit.

Also noted in javadoc: `headerDoesNotRelocateWhenItCannotShareAPageWithTheWholeBody` asserts an *absence* and so reads identically to "the opt-in was never applied" — it pins the documented limit but cannot fail for the right reason on its own. The two sweeps are what drive the behaviour.

Full eight-module reactor `clean verify` — BUILD SUCCESS.

PR #465, opened 2026-07-26
fix(benchmarks): rebuild every branch-versioned module before an A/B run

DemchaAV:
## Why

The A/B scripts installed only the engine module, but `benchmarks/pom.xml` resolves `graph-compose-render-pdf` and `graph-compose-templates` — plus the templates tests-classifier jar, which is deliberately never published — at `${graphcompose.version}`, i.e. the branch's own version.

So both sides of an A/B took those two jars from `~/.m2`. When the compared branches share a version, which is the normal case for two branches cut from `develop`, they took the **same** ones: any change in the PDF backend or the templates module measured a delta of exactly zero. A stale backend jar built against the other branch's core could also throw `NoSuchMethodError` mid-run instead of reporting a number.

## What

`scripts/ab-bench.sh` and `scripts/run-benchmarks.ps1` now install `render-pdf` and `templates` per branch alongside the engine, each guarded by the presence of its pom so a pre-split layout — where the root build produces them — still works. This mirrors the install sequence the CI performance job already uses (`.github/workflows/ci.yml:346-361`).

`graph-compose-fonts` is left alone deliberately: it tracks an independent version line, so it does not change with the branch under test.

## Tests

Both scripts parse clean (`bash -n`, PowerShell AST parser).

The install logic itself was exercised end to end across the layout boundary while investigating engine performance: 8 build-and-measure cycles over `v1.8.0`, `v1.9.1`, `v2.0.0`, `v2.1.0` and `develop`, in a detached worktree. `v1.8.0` predates the module split and has no `render-pdf/` or `templates/` directory, so it takes the guarded-skip path; the 2.x refs take the install path. All eight produced benchmark output.

`scripts/ab-bench.sh` itself was not run end to end here — its own CI smoke (`ab-bench-smoke.yml`) covers that and fires on changes to the script.

### DIFF
--- a/docs/operations/benchmarks.md
+++ b/docs/operations/benchmarks.md
@@ -230,10 +230,12 @@ The wrappers above benchmark whatever is currently checked out. To answer "is
 branch B faster or slower than branch A?" fairly on a noisy laptop, use the
 dedicated A/B scripts. They **interleave** the two branches (A,B,A,B,…) so
 thermal drift averages out, **repeat** each branch and compare **medians**, and
-**cool down** between runs. Each branch is rebuilt (`install -pl :graph-compose-core`) before its
-runs so the benchmark measures that branch's engine, and untracked benchmark
-probes are moved aside around the branch switch so they cannot break the other
-branch's compile.
+**cool down** between runs. Each branch is rebuilt before its runs so the benchmark measures that branch:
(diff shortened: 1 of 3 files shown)


DemchaAV:
The `ab-bench.sh smoke (main vs develop)` check fired on this change and passed, so the script now runs end to end with the added per-branch installs — the verification the description deferred to CI.

PR #497, opened 2026-08-03
docs(examples): the catalogue stops calling current code legacy

DemchaAV:
## Why

Two entries in the example gallery sat under this heading:

> 🗄️ **Legacy** — pre-rebuild examples kept for downstream callers still on V1; do not start new code here. See [`docs/templates/which-template-system.md`](docs/templates/which-template-system.md) for the V1 → V2 migration path.

There is no V1. [docs/templates/which-template-system.md](docs/templates/which-template-system.md) says in its second sentence that GraphCompose 2.0 ships no other template system, both examples compile against `2.1.1-SNAPSHOT` on the current DSL, and three hundred lines below the same file describes one of them as a *"useful starting point when your domain doesn't fit any built-in template"*. A reader looking for exactly that — hand-composed, no preset — was told not to start there.

Separately, the run instructions led with `GenerateAllExamples`. That is not a cheap first command: it is a hand-written list of a hundred `generate()` calls preceded by `ExampleOutputPaths.clean()`, so a reader following the page in order emptied the output directory and rendered the whole catalogue before seeing one document.

## What changed

- The category is now **✍️ Hand-composed** — what those examples are — with the V1 wording and the migration link gone from both the legend and the section heading. The `#handcrafted-proposal` and `#weekly-schedule` sub-anchors are untouched, so nothing that links into the file breaks.
- Run instructions reordered: one example first (`ModuleFirstFileExample`, a Start-here row), then the batch entry point below it, labelled as the way the committed previews and the showcase site get refreshed.
- Release labels the #491 sweep left behind are gone: `## v1.5 feature showcases`, "all v1.8 native vector charts", "the v1.9 book primitives", "byte-identical to v1.4", and "v1.5 cinematic features" in the intro. Also `XxxTemplateV2`, a wrapper name that never existed under that spelling and no longer exists under any.
- [docs/recipes.md](docs/recipes.md): the extension-guide entry no longer dates itself to v1.5, and a 1.x migration guide is dropped from the current cookbook's further-reading list — it remains reachable from the documentation index's *Historical documentation* section.

## What I did not do

The recon for the previous PR noted that four Start-here rows (Engine Showcase, Engine Deck, Twin Output, Maven Central Banner) are bold labels rather than in-page links, unlike the four above them. That is not an oversight: `grep '^### '` shows those four examples have **no detail section** in the file to link to. Giving them anchors means writing four new sections, which is a different change from this one.

## Verification

`./mvnw -B -ntp clean verify` — `BUILD SUCCESS`, exit 0, 692 tests in the closing module. `DocumentationSnippetCompileTest` 6 (this file carries 39 java fences, all still marked), `CanonicalSurfaceGuardTest` 11 (every relative link resolves), `DocumentationCoverageTest` 5.

`grep -n "v1\.[0-9]\|\bV1\b\|Legacy"` over `examples/README.md` returns nothing.

**One thing worth an eyeball rather than a test:** the new pointer `[🚀 Start here](#-start-here)` is the only link in the repository to an emoji heading, and no test validates in-page anchors — `CanonicalSurfaceGuardTest` skips `#` targets by design. The slug follows github-slugger's rule (emoji stripped, leading space becomes the leading hyphen), but the rendered page is the only place to confirm it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -201,6 +201,21 @@ follow semantic versioning; release dates are ISO 8601.
 
 ### Documentation
 
+- **The example catalogue stops calling current code legacy.** Two examples sat under
+  a heading reading "🗄️ Legacy — pre-rebuild examples kept for downstream callers still
+  on V1; do not start new code here", pointing at a V1 → V2 migration guide. There is no
+  V1: that surface was removed in 2.0, both examples compile against the current DSL,
+  and three hundred lines further down the same file called one of them a "useful
+  starting point". They are what they always were — documents assembled from primitives
+  with no template behind them — and the category now says that.
+- **Reading one example no longer starts by rendering a hundred.** The run
+  instructions led with `GenerateAllExamples`, which empties the output directory and
+  regenerates the whole catalogue, and only afterwards mentioned running a single class.
+  A newcomer following the page in order paid for the catalogue before seeing one
+  document. Single-example first now, with the rule for deriving any other example's
+  main class from its source path, and the batch entry point below it — described as
+  what it actually is, the source both `cut-release.ps1` and `ShowcaseSync` read from
+  rather than something that publishes anything itself.
 - **The landing page lets you run it before it finishes selling it.** Ninety-one lines
   and twelve images stood between the title and `## Installation`, most of them a second
   showcase: thirty-five lines demonstrating the PowerPoint backend, with two renders and
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -35,15 +35,12 @@ authoring API; public application code should not import
 For longer-form material:
 
 - [Extension guide](contributing/extension-guide.md) — walkthrough of the four
-  extension paths with the v1.5 `ShapeContainerNode` work as a
-  worked example.
-- [Migration v1.4 → v1.5](roadmaps/migration-v1-4-to-v1-5.md) — every public
-  API change in v1.5 plus suggested migration order.
+  extension paths, with `ShapeContainerNode` as the worked example.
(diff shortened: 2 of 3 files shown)


DemchaAV:
Both fixed, and both deserved the catch — a PR about instruction accuracy that shipped two inaccurate instructions in the run path is the worst place for them.

**1. The class-name rule is now stated and proven.** The text gives the derivation explicitly — drop `src/main/java/` and `.java`, swap `/` for `.` — with the worked example you suggested. Then I checked it generalises rather than assuming: every one of the ~40 source paths linked from this file declares a package matching its directory (zero mismatches), and each Start-here class has a `main`. Finally I ran it end to end on a row I had *not* pasted — derived `com.demcha.examples.templates.invoice.InvoiceCinematicFileExample` by the rule, ran it, exit 0, `Generated: …invoice-cinematic.pdf`, one document.

**2. `GenerateAllExamples` publishes nothing — corrected.** Verified in the source before rewriting: it writes only into `examples/target/generated-pdfs/`; `ShowcaseSync` copies into `web/showcase/`, renders the PNG thumbnails and rewrites `web/examples.json`; and `Refresh-CommittedPreviews` in `cut-release.ps1` walks the generated tree and re-copies the *already committed* subset under `assets/readme/examples/`. So the batch runner is the source both of those read from, which is what the text says now. The same wrong sentence was in the CHANGELOG entry and is fixed there too.

I left the `ShowcaseSync` command out of the block: this section is the reader's path to running an example, and that tool is maintainer-side, driven by the cut. Naming it and what it does seemed the right amount. Say the word if you want the command shown too.

`./mvnw -B -ntp clean verify` — `BUILD SUCCESS`, 692 tests in the closing module.

Head is now `06befc73`.

PR #498, opened 2026-08-03
docs(nav): one recipe catalogue, and the unreachable pages are reachable

DemchaAV:
## Why

Four navigation defects, all of the same shape: a reader cannot get to something, or gets to two versions of it.

**Two recipe catalogues.** [docs/recipes.md](docs/recipes.md) and [docs/recipes/README.md](docs/recipes/README.md) each carried a hand-maintained table of all 22 recipe pages. They agreed on *which* pages exist — and disagreed on the descriptions, in both directions: the folder index was materially richer on 10 rows (it names PDF417 and DataMatrix among the barcode symbologies; the cookbook does not), the cookbook on 8 (themes, transforms, page backgrounds, shape-as-container…). Neither was a subset, so "delete one" would have silently dropped content. The documentation index papered over it by calling one an index and the other a coverage map; they were the same list.

**Unreachable pages.** The archive index named its files in code spans, not links — a table of contents that leads nowhere — and `implementation-guide.md` was not listed at all, which is why an orphan sweep flags it. The two notes beside the committed benchmark baseline, `BASELINE_SUMMARY.md` and `COMPARISON.md`, had no inbound link from anywhere in the repository.

**A decision table with no default.** [docs/getting-started.md](docs/getting-started.md) asked three "yes" questions — known template family, pixel-level control, re-usable new document type. A reader generating an ordinary report from data matches none and falls off the end.

## What changed

- The 22 rows are merged into `docs/recipes.md`, taking the better description of each pair, so nothing is lost. Verified by diffing the two tables key-by-key before and after; 11 rows took the folder's wording, the other 11 kept the cookbook's.
- `docs/recipes/README.md` keeps the one job GitHub gives it — being what renders when you open the directory — and points at the catalogue instead of holding a second copy.
- `docs/README.md` stops describing the pair as two different things.
- The archive index links its three files, and now lists `implementation-guide.md` with what replaced it (the extension guide and the package map).
- `docs/operations/benchmarks.md` names the two baseline notes where it explains the baseline they produced.
- The starting-point table gains the default row: the DSL is the answer, not the fallback.

## One planned item turned out to be already done

The plan's first Phase 6 action was "move `## Scope and comparison` below `## Architecture`". It already is — `## Architecture` at README.md:298, `## Scope and comparison` at :331. That instruction was carried over from an earlier plan written when the order was the other way round, and the reorder landed before this line was ever executed. Recorded rather than acted on; no README change here.

## Verification

`./mvnw -B -ntp clean verify` — `BUILD SUCCESS`, exit 0, 692 tests in the closing module. `CanonicalSurfaceGuardTest` 11 (every relative link resolves), `DocumentationCoverageTest` 5, `DocumentationSnippetCompileTest` 6.

Orphan sweep over `docs/` before and after: `baselines/BASELINE_SUMMARY.md` and `baselines/COMPARISON.md` go from unreferenced to linked. The four files still reported are the three folder `README.md`s — reached by opening the directory, which a link scanner cannot see — and the archived pages, now linked from their own index.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -201,6 +201,26 @@ follow semantic versioning; release dates are ISO 8601.
 
 ### Documentation
 
+- **One recipe catalogue instead of two.** The cookbook page and the folder index each
+  carried a hand-maintained table of all twenty-two recipe pages. They happened to agree
+  on which pages exist and disagreed on ten of the descriptions — the folder index named
+  PDF417 and DataMatrix among the barcode symbologies, the cookbook did not; the cookbook
+  was fuller on themes, transforms and page backgrounds. Merged into the page both the
+  README and the documentation index point at, taking the better description of each
+  pair, so nothing is lost. The folder file keeps the one job GitHub gives it — being
+  what you see when you open the directory — and now says where the catalogue is instead
+  of holding a copy that drifts.
+- **Documents nobody could reach are reachable.** The archive index named its three
+  files in code spans rather than links, so the folder's own table of contents did not
+  actually lead anywhere, and one of the three was not listed at all. The two benchmark
+  notes sitting beside the committed baseline had no inbound link from anywhere; the
+  benchmark guide names them now, and says what they are — an April 2026 capture and the
+  before/after of the optimization round it belonged to, six weeks older than the
+  baseline file and unrelated to it.
+- **The starting-point table has an answer for the ordinary document.** Its three rows
+  each asked a "yes" question — a known template family, pixel-level control, a re-usable
+  new type — and a reader generating an ordinary report from data matched none of them
+  and fell off the end. The DSL is the default, not the fallback, and the table says so.
 - **The example catalogue stops calling current code legacy.** Two examples sat under
   a heading reading "🗄️ Legacy — pre-rebuild examples kept for downstream callers still
   on V1; do not start new code here", pointing at a V1 → V2 migration guide. There is no
--- a/docs/README.md
+++ b/docs/README.md
@@ -79,7 +79,7 @@ Kept for anyone stepping through the 1.x line one minor at a time. Nothing here 
 </details>
 
 ### Recipes (cookbook-style howtos)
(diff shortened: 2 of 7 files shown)


DemchaAV:
You are right, and I verified it rather than taking the account on trust — git says exactly what you said:

- `baselines/BASELINE_SUMMARY.md` and `COMPARISON.md` were both added in `2f63c3dd`, **2026-04-27** ("Bump to v1.3.0 and refresh release docs").
- `baselines/current-speed-full.json` first appears in `b8267845`, **2026-06-08** — six weeks later, in "perf(bench): add current-speed verdict gate" — then re-cut in `4c0dee0d` (2026-06-09) and `fbc23cfb` (2026-07-27).

So the April notes predate the baseline by six weeks and produced nothing that feeds it. I read the two files' titles and inferred a lineage instead of checking. That is the failure mode this whole series of PRs has been fixing in other people's prose, which makes it a poor one to commit.

**Corrected.** The paragraph now says the notes are historical context only, dates the April capture, and states plainly that neither produced the baseline nor tracks it — the JSON was cut separately when the verdict gate landed and has been re-measured independently since. It also now tells the reader to read the JSON's own `timestamp` / `profile` / `sourceRuns` fields rather than infer provenance, which is the durable fix.

I also moved the aside out of the middle of the refresh procedure: my first attempt dropped it between "this is the baseline" and "refresh it only for a verified improvement", splitting one instruction in half. It now sits with the provenance sentence at the end of that section.

The CHANGELOG carried the same invented link — "the before/after it was ratcheted against" — and is fixed too. The PR body above is stale on this point; treat this comment as the correction.

**On the non-blocking note:** a guard asserting every `docs/recipes/*.md` appears in the `docs/recipes.md` table is worth having, and it is now the natural pair to the anchor-validation guard already queued from #495. Both are on the follow-up list rather than in this PR.

`./mvnw -B -ntp clean verify` — `BUILD SUCCESS`, 692 tests in the closing module. Head is now `3a37c95e`.

PR #502, opened 2026-08-04
docs(charts): the recipe documents the styling surface it has

DemchaAV:
## Why

The chart recipe is the only page describing chart styling, and it had drifted from
the builder in three ways.

Five settings appeared nowhere on it — `axisTextStyle`, `legendTextStyle`,
`valueLabelTextStyle`, `donutCenterTextStyle`, `barWidthRatio` — though four are
load-bearing in the flagship examples that put a chart on a dark card. The only way
to find them was to read `ChartStyle.Builder`.

`valueLabelHalo` was documented under line charts, the one place its default white
chip is least likely to be wrong, rather than as the backing behind value and slice
labels alike. The single setting that must change when a chart leaves a white page
was filed where a bar or donut author would not look.

And `ChartTheme` was presented as a layer an author styles through. A chart placed
through the DSL resolves its geometry during the layout pass, after the document's
theme is out of reach (`ChartDefinition.java:67`), so it always starts from
`ChartDefaults.DEFAULT_THEME` and `ChartStyle` is the author-facing override. The
low-level `ChartLayoutResolver.resolve(...)` does take an explicit `ChartTheme` —
the page now draws that line rather than implying either that the theme is a knob
or that nothing accepts one.

## What

- Every `ChartStyle` setting in one table with its default and what it applies to,
  marking which are inherited from `ChartDefaults.DEFAULT_THEME` and which are fixed
  engine defaults.
- A section for the halo of its own: what it is for, why white is wrong off a white
  page, and translucency through `withOpacity(...)`.
- A typography section for the three chrome text styles; `donutCenterTextStyle`
  documented beside the `centerText` it styles; `barWidthRatio` in the bar section.
- `ChartStyleDocumentationGuardTest` reflects over `ChartStyle.Builder` and fails the
  build when a setter is not named on the page.

## Tests

- `./mvnw -B -ntp clean verify` → `BUILD SUCCESS`.
- The guard was confirmed able to fail: with `barWidthRatio` and `donutCenterTextStyle`
  removed from the page it reported exactly those two.
- The three new snippets carry `doc-example` markers, so `DocumentationSnippetCompileTest`
  compiles them against the current API; breaking `valueLabelHalo` produced a compiler
  error attributed to `charts-halo`.

## Base

Rebased onto `develop` at #501, which is what makes the halo section true for every
chart kind: grouped-bar value labels passed `null` until that fix, so the page would
otherwise have promised behaviour the default bar chart did not have.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -211,6 +211,21 @@ follow semantic versioning; release dates are ISO 8601.
 
 ### Documentation
 
+- **The chart recipe documents the styling surface it has.** Five settings were named
+  nowhere on the page — the three text styles, the donut centre style and the bar width
+  ratio — four of them already load-bearing in the flagship examples that put a chart on
+  a dark card, and discoverable only by reading the builder's source. The
+  value-label halo was filed under line charts, the one place its default white chip is
+  least likely to be wrong, rather than described as what it is: the backing behind
+  value and slice labels alike, and the first thing that must move when a chart leaves a
+  white page. And `ChartTheme` was presented as a layer an author styles through, though
+  no authoring API accepts one — a chart resolves its geometry after the document's theme
+  is out of reach, so `ChartStyle` is the author-facing whole of it. The low-level
+  `ChartLayoutResolver.resolve(...)` does take a `ChartTheme`, and the page now says so
+  rather than leaving it out. The page carries every setting
+  with its default and what it affects, worked examples for typography and for the halo,
+  and `ChartStyleDocumentationGuardTest` fails the build when a setter reaches the builder
+  without reaching the page.
 - **One recipe catalogue instead of two.** The cookbook page and the folder index each
   carried a hand-maintained table of all twenty-two recipe pages. They happened to agree
   on which pages exist and disagreed on ten of the descriptions — the folder index named
--- a/core/src/test/java/com/demcha/documentation/ChartStyleDocumentationGuardTest.java
+++ b/core/src/test/java/com/demcha/documentation/ChartStyleDocumentationGuardTest.java
@@ -1,0 +1,75 @@
+package com.demcha.documentation;
+
+import com.demcha.compose.document.chart.ChartStyle;
+import org.junit.jupiter.api.Test;
+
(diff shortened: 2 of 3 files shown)


DemchaAV:
All three are real, and the branch is now rebased on `develop` at #501, so the halo section is unconditionally true — both `BarChartLayout` call sites pass `halo`.

**1. `ChartTheme`.** You are right and my claim was simply too wide. `ChartLayoutResolver` is a `public final class`, `resolve(...)` is `public static` with no `@Internal`, it takes a `ChartTheme`, and `ChartTextMetrics` is a public interface — so the seam is genuinely callable, not theoretical. The page now separates the authoring path (always `ChartDefaults.DEFAULT_THEME`, `ChartStyle` is the override) from the geometry seam that does accept a theme; CHANGELOG says "no authoring API accepts one".

This one had a second half neither of us named: the layers table at the top called `ChartLayoutResolver` an *"internal pure function `(data, spec, style) → primitives`"* — internal for something public, and a signature that does not exist (the real one is `(spec, style, theme, width, height, metrics)`). Fixed in the same pass, since it is the same error one layer up.

**2. Footnote.** Confirmed: `lineWidth` and `valueLabelOffset` live in package-private `ChartLayoutSupport`, `areaOpacity` is private in `LineChartLayout`, `sliceGapDegrees` is a literal in `PieChartLayout`, and `pointMarker` / square corners are the absence of an override rather than any constant. Reworded to your form, with the nuance kept that three unmarked rows (`barWidthRatio`, `sliceStroke`, `donutCenterTextStyle`) *are* `ChartDefaults` constants.

**3. `valueLabelOffset` on pie/donut.** Confirmed at `PieChartLayout:61` and `:114` — `valueLabelGap(style)` sets the radial gap between the rim and the slice label. Fixed in both the table row and the prose.

Worth noting where the guard's limit is: it checks that each setter is *named* on the page, so it would not have caught any of these three. They are claims about behaviour, not missing names — that part still rests on review.

`clean verify` → `BUILD SUCCESS`. Amended rather than stacked a fix commit, so the branch is still one commit.

PR #505, opened 2026-08-04
chore(build): the gates cover the surfaces they claim

DemchaAV:
## Why

Three gates reported on less than their names implied. Each failure mode is silence:
the step goes green having checked less than the reader assumes.

**The Javadoc gate skipped the entry point.** It ran with
`<subpackages>com.demcha.compose.document</subpackages>`, so `GraphCompose` — in the
root package, and the class every README snippet starts from — was never linted.
`GraphCompose.html` was not even generated by the report goal. It had carried a real
doclint error since `e1376ad8` (the layout move, 2026-07-12): two `<h3>` headings under
an implicit `<h1>`, exactly the sequence break `doclint` exists to catch. The other
path could not catch it either — the published Javadoc jar is built with `doclint=none`
so a broken tag never blocks a release.

**The CI gate-coverage guard could not see half the job names GitHub allows.** Its job
pattern was `^  ([a-z][a-z0-9-]*):$`, which admits what today's names happen to use. A
job added as `build_and_test` or `CodeQL` matched nothing, and a job the guard never
sees is one it cannot report missing from `ci-gate.needs` — the same silence the guard
was written to break.

**Nothing held `docs/recipes.md` to the folder it indexes.** It is the page the README
and the documentation index both point at, so a recipe missing from it has no inbound
link from anywhere, and adding one disturbs no gate.

## What

- `core/pom.xml`: the Javadoc gate's scope becomes `com.demcha.compose`; the two
  headings in `GraphCompose` become `<h2>`.
- `CiGateCoverageGuardTest`: job names are taken structurally, from the YAML's own
  indentation, rather than from a guess at GitHub's identifier grammar.
- `RecipeCatalogueGuardTest` (new): fails on a page the catalogue omits, and on a
  catalogue row pointing at a file that is gone.

Widening the Javadoc scope also pulls in the `@Internal` `com.demcha.compose.engine`
package. That is a consequence, not a goal: with `subpackages` set,
`excludePackageNames` does not take effect — both `a.b.engine:a.b.engine.*` and
`a.b.engine*` were tried and the package is emitted either way. It costs warnings, not
failures, since `failOnError` fails on errors only. Recorded in the pom rather than left
for the next person to rediscover.

## Tests

`./mvnw -B -ntp clean verify` → `BUILD SUCCESS`; `./mvnw -B -ntp javadoc:javadoc -pl
:graph-compose-core` → exit 0 (it exited 1 on the widened scope before the heading fix).

Each guard was checked from both sides rather than trusted because it was green:

- **Job pattern.** Appending a `security_scan:` job absent from `ci-gate.needs` makes
  the old pattern pass — it does not see the job at all — and the new one fail, naming
  `security_scan`.
- **Recipe catalogue.** An unlisted page is named (`probe-page.md`); a catalogue row
  pointed at a renamed file is reported from both ends.

The published Javadoc jar is unchanged: built with `-Prelease` before and after, both
carry 1024 entries and the same package set, so the widened `subpackages` reaches the
report goal only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,36 @@ follow semantic versioning; release dates are ISO 8601.
 
 ### Build
 
+- **The Javadoc gate lints the class readers open first.** It ran with
+  `subpackages` set to `com.demcha.compose.document`, so `GraphCompose` — the entry
+  point every snippet in the README starts from — sat in the root package outside it,
+  and had done since the module layout moved. It carried a real doclint error the
+  whole time: two `<h3>` headings under an implicit `<h1>`, which is the sequence
+  break `doclint` exists to catch. Nothing else caught it either, because the
+  published Javadoc jar is built with `doclint=none` so a broken tag never blocks a
+  release. The gate now covers the whole of `com.demcha.compose`, root package
+  included, and the two headings are `<h2>`. Widening it also pulls in the
+  `@Internal` engine package: `excludePackageNames` does not take effect alongside
+  `subpackages`, and that costs warnings rather than failures, since `failOnError`
+  fails on errors only.
+- **A CI job the gate guard cannot name is a job it cannot report.** The guard that
+  checks every pull-request job is aggregated by `CI Gate` found those jobs with a
+  pattern admitting lower case and hyphens — everything today's names happen to use.
+  A job added as `build_and_test` or `CodeQL` was not matched, and neither was
+  `security_scan: # nightly` or a name with a space after the colon, because the
+  pattern also required the line to end there. A job the guard never sees is one it
+  cannot report missing: it sits outside the gate's `needs`, the test stays green, and
+  the aggregate check branch protection requires is blind to it. Job names are now
+  taken structurally, from the YAML's own indentation, rather than from a guess at
+  GitHub's identifier grammar, and the parser takes the workflow as text so
+  `CiGateCoverageGuardParsingTest` can drive it with the job spellings this
+  repository's own workflow does not happen to contain.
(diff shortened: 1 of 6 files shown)


DemchaAV:
Confirmed and fixed. Reproduced the blind spot first — appending

```yaml
  security_scan: # Run security tools
    runs-on: ubuntu-latest
```

to `ci.yml`, with the job absent from `ci-gate.needs`, left the guard **green**. Exactly the silence the PR set out to remove, one shape further along than I had checked.

**Pattern.** Now `^  ([^\s:#]+):[ \t]*(?:#.*)?$`. Beyond the inline comment this also covers trailing whitespace after the colon, which the old pattern dropped the same way — found while writing the cases, not reported.

**Testability, rather than one more fixture.** The guard reads `.github/workflows/ci.yml`, so it can only ever exercise the spellings that file happens to use; every other one stays untested until a real job lands on it. `jobBlocks` now takes the workflow as text and is package-private, and `CiGateCoverageGuardParsingTest` drives it with five shapes the repository's workflow does not contain: underscore and capitalised ids, an inline comment, trailing whitespace, nested keys that must *not* be read as jobs, and a job block that must not inherit the previous job's `if:`. Reverted to the old pattern, two of those five fail — so they test the fix rather than describe it.

**Recipe link grammar.** Taken as you suggested: `[^/#)\s]+\.md` instead of the lower-case-and-hyphens spelling. Agreed it was a safe failure, but a guard that cries wolf over a legal file name is one the next person edits until it stops complaining.

Rebased onto `develop` at #506 (`33731d4c`). `clean verify` → `BUILD SUCCESS`; `javadoc:javadoc -pl :graph-compose-core` → exit 0. Amended rather than stacked, so the branch is still one commit.

Issue #528, opened 2026-08-09
DOCX: a table's fill and borders are not painted


DemchaAV:
`DocumentTableStyle` carries a fill and a border spec, and `DocxSemanticBackend.writeTable`
applies neither. A table exports with the right structure — `colSpan`/`rowSpan` reach Word as
`w:gridSpan`/`w:vMerge` since #524 — and with the cascaded text style on its cells, but a
zebra fill, a header band, or a per-cell border comes out on Word's defaults.

Word expresses both: `w:tcPr/w:shd` for the cell fill and `w:tcPr/w:tcBorders` for the edges,
and the cascade to resolve them from (table → column → row → cell) is already computed by
`resolveCellTextStyle` for the text half.

Not in #447: that issue covered data the node carries and the backend discards on the *text*
side, and its five fix directions are done. This is the paint that was never attempted.

**Also update** `docs/architecture/backend-capability-matrix.md` (the table row names this as
what is still missing) and `render-docx/README.md`.

DemchaAV:
Done on `develop` (`b862b0bf`), by the work that landed alongside the table structure rather than by a change referencing this issue — which is why it stayed open.

**The paint.** `DocxSemanticBackend.applyCellPaint` maps a cell's fill to `w:tcPr/w:shd` (`STShd.CLEAR` + the hex fill) and its stroke to `w:tcPr/w:tcBorders`, writing all four edges. Border width goes out in the eighths of a point `w:sz` counts, rounded up to at least 1 so a hairline the author asked for stays a line instead of disappearing. A stroke of zero width is written as "no border" rather than omitted, so it overrides an inherited edge.

**The cascade** resolves per field, not per style — a row that sets only a fill keeps the table's stroke, matching what `resolveCellTextStyle` already did for the text half.

**What does not survive**, recorded rather than silently dropped: `w:shd` is opaque, so a fill carrying an opacity below 1 lands at full strength. Blending it would need a background this backend does not resolve, Word owning the flow.

**Tests** — `DocxTablePaintTest`, 6 cases: fill reaches Word as shading, a stroke becomes four borders in eighths, the cascade resolves per field, a merged cell is painted on every position it covers, a hairline stays a line, and a zero-width stroke says "no border".

**Both documents the issue asks for are updated**: the capability matrix's table row now names `w:shd` / `w:tcBorders` and the opacity limit, and `render-docx/README.md` describes the same in its table section.

PR #532, opened 2026-08-09
fix(docx): paint a table the way it was styled

DemchaAV:
## Why

`DocumentTableStyle` carries a `fillColor` and a `stroke`, and `writeTable` read neither. A
zebra body, a header band and a ruled grid all exported on Word's defaults — which is to say
with no fill, and with whatever borders POI puts on a table it creates. Word owns both
(`w:shd`, `w:tcBorders`), so nothing here is an approximation of a thing Word cannot do.

## What changed

- **Fill maps to `w:shd`, stroke to `w:tcBorders`** — four edges, `w:sz` in eighths of a point
  rounded up to at least one so a hairline the author asked for stays a line.
- **The cascade resolves per field.** `resolveCellTextStyle` became
  `resolveCellValue(node, placement, field)`, so the table → column → row → cell order now
  applies to each property on its own: a table-wide rule survives a row that only overrides the
  fill. That matches how `TableLayoutSupport.buildStylesGrid` merges for the fixed-layout path.
- **A merged cell is painted on every position it covers.** A `w:vMerge` continuation draws its
  own shading, so leaving it unpainted would stripe the region.
- **A stroke of no width says so.** `DocumentStroke.of(colour, 0)` is this codebase's "no
  border" — `PdfTableRowFragmentRenderHandler` reads the same predicate as draw-nothing, and
  the shipped `MintEditorial` CV preset uses it. Writing nothing left the cell on the grid Word
  puts on a table by default, so a deliberately borderless design exported ruled. The cell now
  emits `STBorder.NIL`. A table that says *nothing* about borders still keeps Word's default:
  the look belongs to Word where the document did not claim it.

What a fill loses is its opacity. `w:shd` is opaque, and blending it needs a background this
backend does not resolve — Word owning the flow is the whole premise of the semantic export.

**The example carries the feature.** The catalogue had no DOCX example with a styled table, so
this would have shipped with no render behind it — `WordExportExample`'s table now takes a
header band and a table-wide rule. Both of its committed previews move: the DOCX for the new
markup, and the PDF because the fixed-layout backend paints a style it was never given before.
That second one is the drift guard doing exactly what it was widened for last week.

## Verification

`./mvnw -B -ntp clean verify -pl :graph-compose-core,…,:graph-compose-coverage -am` →
**BUILD SUCCESS**; `./mvnw -B -ntp test -f examples/pom.xml` → **BUILD SUCCESS**, 73 tests, and
no preview drift once both were refreshed.

New `DocxTablePaintTest`, **6 tests**: a fill reaching `w:shd` while an unpainted cell keeps
Word's default; a stroke becoming four `SINGLE` edges at the right `w:sz` and colour; the
cascade resolving fill and stroke from different levels at once; a merged cell painted on both
its positions; a hairline that rounds to one eighth rather than to nothing; and a zero-width
stroke emitting `NIL`.

Reverting `DocxSemanticBackend.java` to `develop` fails all six.

The emitted `CT_TcPr` was checked against the schema by unzipping a document that carries a
span *and* paint on one cell: the order comes out `w:gridSpan → w:vMerge → w:tcBorders →
w:shd`, the `CT_TcPrBase` sequence, because XmlBeans inserts each child at its schema position
rather than in call order.

## Documentation

`docs/architecture/backend-capability-matrix.md` and `render-docx/README.md` name what the
paint now carries and what it drops. The root `README.md` DOCX row said fill and border paint
were dropped; that line is the one this series keeps having to correct, and it is corrected
again here.

## Not covered

Borders on a `w:vMerge` continuation are written on all four edges, so a merged region
formally carries interior rules. Word is expected to suppress them; I could not pin that in a
test, so it is named rather than claimed.

**Lane:** shared-engine — a render backend. No public API change.

Closes #528.


DemchaAV (commented):
*-*

Issue #140, opened 2026-06-08
Arabic, Hebrew, other Unicode support


jottinger:
From Reddit: does it support Arabic? This creates other questions: Hebrew (and other left-to-right languages).

DemchaAV:
**Baseline landed.** #141 (merged) adds a multilingual smoke test proving GraphCompose **renders** Hebrew/Arabic/Hungarian glyphs to a valid PDF (cross-platform: Windows/macOS/Linux). It does **not** make RTL correct yet.

Keeping this issue open to track **full Arabic/Hebrew (RTL) support** — the engine currently has no BiDi reordering and no Arabic shaping (text is drawn in logical order):

- [ ] RTL paragraph direction (explicit + auto base-direction)
- [ ] Right alignment for RTL paragraphs/blocks
- [ ] BiDi reordering (UAX #9) for mixed Arabic/Hebrew + Latin + digits (`java.text.Bidi` or ICU4J)
- [ ] Arabic shaping / contextual forms (initial/medial/final/isolated, lam-alef)
- [ ] RTL-aware line wrapping
- [ ] Punctuation / bracket mirroring in RTL
- [ ] Snapshot/visual tests with known-correct output, verified by a native-script reader

**Acceptance:** a mixed Arabic/Hebrew + English + numbers paragraph renders with correct base direction, BiDi order, Arabic joining, and right alignment.

DemchaAV:
Right-to-left support is implemented and open for review as a four-PR stack, to be merged bottom-up:

| PR | What |
|---|---|
| #535 | Amiri (Arabic) and David Libre (Hebrew) in `graph-compose-fonts` 1.1.0 |
| #534 | the families in the catalog as `FontName.AMIRI` / `DAVID_LIBRE`; engine line moves to 2.2.0 |
| #536 | `TextDirection` (LTR / RTL / AUTO), UAX #9 resolution, visual ordering across all three backends |
| #538 | Arabic contextual joining and paired-punctuation mirroring |

Against the acceptance criteria in the description:

- **Base direction** — `ParagraphBuilder.direction(...)`; `AUTO` resolves from the paragraph's first strong character, once per paragraph as UAX #9 P2–P3 requires.
- **Bidirectional order** — lines are split into single-direction spans and carry a visual-order permutation; embedded Latin and digits keep running forwards inside right-to-left text. Proven by reading glyph X-positions out of the rendered PDF, not by inspecting the model.
[...]

PR #534, opened 2026-08-10
feat(fonts): register the Arabic and Hebrew families in the catalog

DemchaAV:
Stacked on #535 — that PR carries the font binaries and the `graph-compose-fonts` 1.0.0 → 1.1.0 bump; this one is the engine side. Review after it.

## Why

`graph-compose-fonts` 1.1.0 carries Amiri and David Libre, but nothing selects them: the catalog is a hand-written list of `FontName` constants, so a bundled family that is not registered is unreachable except through `FontFamilyDefinition`.

## What changed

`FontName.AMIRI` and `FontName.DAVID_LIBRE` join `DefaultFonts`, and `graphcompose.fonts.version` moves to 1.1.0 in every pom that pins it.

`DefaultFonts.regularAndBold(...)` is the catalog entry for a family with fewer than four faces — David Libre ships no italic upstream, so `FontFamilyDefinition.build()` collapses italic onto regular and bold-italic onto bold.

`FontFamilyDefinition` now tells a stale font artifact apart from a missing one. Both used to produce the same message — an instruction to add a dependency that was already on the classpath. It probes a face that has shipped since the v1.8.0 split: artifact present means the family is newer than the artifact, and the message names the version carrying it; artifact absent keeps the original migration pointer.

The engine line moves to `2.2.0-SNAPSHOT`, and the open changelog entry with it, because this adds public API.

**Scope:** this makes the glyphs selectable. Text is still laid out in logical order — the bidirectional ordering and Arabic joining that use these forms are the next steps on #140.

## Verification

Full reactor gate green:

```
./mvnw -B -ntp clean verify -pl :graph-compose-core,:graph-compose-render-pdf,:graph-compose-render-docx,:graph-compose-render-pptx,:graph-compose-templates,:graph-compose-testing,:graph-compose-qa,:graph-compose-coverage -am
```

`BUILD SUCCESS` across all 11 modules including `coverage`.

- `ArabicHebrewFontCoverageTest` (render-pdf, 4 tests) — every code point in U+FE80–FEFC encodes in all four Amiri faces, the Hebrew block in both David Libre faces, Latin in both families, and David Libre's italic resolves to its regular program. This is the guard that makes a swap to a `GSUB`-only family fail rather than silently unjoin Arabic.
- `BundledRtlFamiliesTest` (core, 5 tests) — both families are in `bundledFontNames()`; David Libre's face fallbacks are pinned; Amiri's four faces are asserted distinct, since a copy-paste collapsing one of them would pass every coverage assertion; and each missing-resource message is asserted on the signal it carries rather than on a marker that has to stay absent.
- `ArabicHebrewGlyphsDemoTest` (qa, 2 tests) — renders each script through its family to `target/visual-tests/rtl-fonts/` and asserts the extracted text contains no `?`, which is what a family that cannot encode the script produces.
- `MeasurementFontParityTest` covers 32 families instead of 30, so measured and painted widths stay in lockstep for the new binaries.

`available_fonts_preview` moves because that document renders the whole catalog: 364 added lines are the two new sections, 4 removed lines are the enclosing container's height and offset. No existing family's geometry changed.

**Merge order:** CI here cannot pass until `fonts-v1.1.0` is published. Several jobs build a single module without `-am` and resolve `graph-compose-fonts` from Central, so the 1.1.0 pin fails to resolve until the tag exists — which is the order [the release process documents](https://github.com/DemchaAV/GraphCompose/blob/develop/docs/contributing/release-process.md): tag the fonts release, then bump the property.


DemchaAV:
Sequencing note, since this PR's description still reads as if `fonts-v1.1.0` goes up
before the rest of the stack.

It must not. `graph-compose-fonts:1.1.0` is unreleased — `fonts-v1.1.0` is not tagged
(only `fonts-v1.0.0` exists) and Central returns 404 for it — and #539 adds Georgian,
Armenian and Korean to that same version. A Central artifact is immutable, so a tag cut
between the font PRs would publish a 1.1.0 that the catalog then tells consumers contains
families it does not, failing as a missing resource at first use rather than at build time.

The order is: merge the stack through #539, then cut one `fonts-v1.1.0`. If the tag has
already gone up by the time this lands, #539 moves to 1.2.0 instead and every pin and
minimum-version entry moves with it.

The rule itself is now written down in `docs/contributing/release-process.md` rather than
living in a PR description, along with the correction that six poms re-pin the fonts
version rather than two, and that `core` does carry the pin at test scope — so a
single-module build needs the artifact installed first.

PR #535, opened 2026-08-10
feat(fonts): add Arabic and Hebrew families to the bundled set

DemchaAV:
## Why

The bundled set had no family covering Arabic or Hebrew, so both scripts rendered as `?` unless the caller registered a font of their own. The multilingual smoke test does exactly that, and skips itself when the host has no suitable font.

Picking the Arabic family is not only a coverage question. A PDF draws text through the font's `cmap` and does not execute OpenType `GSUB`, so contextual Arabic letter forms are only ever reachable when the font itself carries the Arabic presentation forms. Scheherazade New and Rubik cover the Arabic block completely and carry none of them — they render, measure, and produce a valid PDF with every letter drawn isolated, which nothing in the output announces. **Amiri** carries them; that is why it was chosen over the alternatives.

## What changed

- `fonts/src/main/resources/fonts/google/amiri/` — four faces plus `OFL.txt`.
- `fonts/src/main/resources/fonts/google/davidlibre/` — regular and bold plus `OFL.txt`; upstream ships no italic.
- `fonts/pom.xml` 1.0.0 → **1.1.0**, with the install snippets in `fonts/README.md` following it (the version guard holds those in step).

Neither family covers the other's script, so a run mixing Arabic and Hebrew still needs a font of your own. `fonts/README.md` says so.

Packaged jar: 16.6 MB → 17.5 MB.

## Verification

Both licences are the SIL OFL 1.1 shipped by each upstream project. Glyph coverage was checked by parsing each face's `cmap`: all four Amiri faces encode U+0621–064A, U+FE80–FEFC and the lam-alef ligatures; both David Libre faces encode U+05D0–05EA. Latin is present in every face, which matters because a run is drawn in a single font and the engine does not fall back across families.

`./mvnw -B -ntp -pl :graph-compose-fonts package` succeeds; the module is resource-only and has no tests of its own.

The engine-side change that registers these as `FontName` constants — with the coverage guard that pins the presentation-form requirement — follows in a separate PR. It has to, in the order [the release process documents](https://github.com/DemchaAV/GraphCompose/blob/develop/docs/contributing/release-process.md): `graphcompose.fonts.version` can only move to 1.1.0 once a `fonts-v1.1.0` tag has published it, because several CI jobs build a single module and resolve the artifact from Central rather than from the reactor.


DemchaAV:
Sequencing note, since this PR's description still reads as if `fonts-v1.1.0` goes up
before the rest of the stack.

It must not. `graph-compose-fonts:1.1.0` is unreleased — `fonts-v1.1.0` is not tagged
(only `fonts-v1.0.0` exists) and Central returns 404 for it — and #539 adds Georgian,
Armenian and Korean to that same version. A Central artifact is immutable, so a tag cut
between the font PRs would publish a 1.1.0 that the catalog then tells consumers contains
families it does not, failing as a missing resource at first use rather than at build time.

The order is: merge the stack through #539, then cut one `fonts-v1.1.0`. If the tag has
already gone up by the time this lands, #539 moves to 1.2.0 instead and every pin and
minimum-version entry moves with it.

The rule itself is now written down in `docs/contributing/release-process.md` rather than
living in a PR description, along with the correction that six poms re-pin the fonts
version rather than two, and that `core` does carry the pin at test scope — so a
single-module build needs the artifact installed first.

PR #536, opened 2026-08-10
feat(engine): lay out and draw right-to-left text

DemchaAV:
Stacked on #534 (which is stacked on #535) — review and merge in that order. This is the engine work of [#140](https://github.com/DemchaAV/GraphCompose/issues/140).

## Why

The engine laid text out strictly left to right: tokens append in source order and the PDF handler advances one cursor from the line's left edge. Text is stored in logical order — the order it is read — and a page draws in visual order. For Latin the two coincide; for Hebrew and Arabic they do not, so every right-to-left line rendered reversed, in a document that otherwise looked finished.

## What changed

`ParagraphBuilder.direction(...)` takes `TextDirection.LTR`, `RTL`, or `AUTO` (read off the paragraph's first strong character, resolved once per paragraph as UAX #9 P2–P3 requires). Direction is a separate choice from `TextAlign` — alignment says where a line sits, direction says which way it runs — and they meet in one place: a right-to-left paragraph aligns right unless the caller chose an alignment, which the builder tracks explicitly because a field defaulted to `LEFT` cannot tell "not chosen" from "chose LEFT".

Resolution lives in a new engine package, `com.demcha.compose.engine.text.bidi`, wrapping `java.text.Bidi` (no new dependency). All three wrap paths carry it: a plain line is split into one span per directional run; inline runs and markdown bodies, already one span per word, are split further wherever the level changes *inside* a token — `"ב-2026"`, one whitespace token, would otherwise reverse its digits. Spans stay in logical order and each line carries a `visualOrder` permutation, validated as one by `ParagraphLine`.

Each backend then does only what its medium requires:

| Backend | Span order | Reverses text | Why |
|---|---|---|---|
| PDF | visual | yes, by grapheme cluster | `showText` draws characters in the order given |
| PPTX | visual | no | PowerPoint's own bidi engine reverses; absolute per-span frames stop it reflowing the word order |
| DOCX | logical | no | Word lays out itself and only needs `w:pPr/w:bidi` |

The bidirectional format characters (LRM, RLM, ALM, embeddings, isolates) now survive control-character sanitizing until the algorithm has read them, and are dropped at the glyph seam both measurement and rendering share — substituting them would have printed `?` and given a zero-width character a width.

A left-to-right line pays nothing: an allocation-free scan (checked against `java.text.Bidi.requiresBidi` by test — it may over-include, it can never miss) bails out before any resolver work, and the LTR branch is the byte-identical code it always was.

## Verification

Full reactor gate green; no committed layout snapshot or visual baseline moved, which is the executable form of the LTR-identity claim.

The load-bearing proofs read glyph X-positions out of rendered pages (`RtlGlyphOrderTest`) — a layout snapshot records node geometry and cannot see span order. They pin: Hebrew drawn right-to-left under both paragraph directions, Latin untouched, an embedded Latin word running forwards inside a reversed line, the digits of a mixed token staying forwards, and an AUTO paragraph whose Latin-initial second line keeps its sibling's base. ~70 new tests across core, render-pdf, render-pptx, render-docx and qa, including grapheme-cluster reversal held against a naive `StringBuilder.reverse` (pointed Hebrew, emoji), the sanitizer's three direction-mark operations, and the PPTX shared-frame/absolute-frame split.

Performance was measured, not argued: the first A/B against this branch's base caught a real regression (+17.7% proposal-template latency from a per-line `char[]` copy and run-record allocation), which is why the scan above exists; the re-run after the fix and a third run after review fixes show every scenario inside the harness's declared noise band, with signs flipping between runs — noise, not effect.

## Known limits

- Arabic renders in unjoined letter forms; contextual shaping via the presentation forms Amiri carries is the next step on #140.
- Plain-text extraction and copy-paste from a PDF read the content stream, which carries the visual order — a Hebrew line pastes reversed until `ActualText` is written. DOCX is unaffected. Recorded in the changelog.
- A run is drawn in one font and no bundled family covers both Arabic and Hebrew; mixing them in one run needs a user-registered font.
- A highlight chip is one unsplittable fill, so a chip whose text changes direction inside itself keeps logical order within the chip.


DemchaAV:
Pushed `20c7bac1`: the direction a paragraph runs in is now read in one place.

**It was read in two, and they disagreed.** The builder scanned for the first character
of strong directionality to decide which edge an unaligned paragraph sits at; the layout
asked the bidirectional algorithm. They agree on ordinary text and part company on
**isolates** — UAX #9 rule P2 skips everything between an isolate initiator and its
matching PDI when looking for the first strong character, which is what an isolate is
*for*, and a plain scan reads straight into it:

| text | builder | layout |
| --- | --- | --- |
| `שלום` | RTL | RTL |
| `2026 שלום` | RTL | RTL |
| `LRI שלום PDI hello` | **RTL** | **LTR** |
| `RLI hello PDI שלום` | **LTR** | **RTL** |

So a paragraph opening with an isolated Hebrew quotation was aligned to the right edge
and laid out from the left. This is the release that started carrying isolates through
control sanitizing, so the disagreement was reachable by exactly the author who took the
documentation's advice on steering a neutral stretch of text.

`ParagraphDirection` is that single answer. It lives in `document.layout` because that is
the sanctioned bridge from the canonical surface to the engine — the builder may call
there, and only there does the call reach `BidiParagraphResolver`. No cycle: nothing in
`document.layout` imports `document.dsl`.

**Word is the third caller.** `AUTO` reached the DOCX export unresolved, on the reasoning
that it had already become a concrete alignment — but alignment is not direction, and
what Word actually received was a paragraph with no base direction at all, left to infer
one. That is what `w:bidi` exists to prevent, and it comes out the other way round for a
line opening with a digit or a parenthesis. The export is marked from the same answer the
page used.

Tests: `ParagraphDirectionTest` covers explicit directions, first-strong with neutrals,
the two isolate cases above, and — the one that matters — that the alignment a paragraph
is *built* with matches the direction it is *laid out* with, across all five inputs. I
checked it fails against the old scan before keeping it; it does, naming
`<LRI>שלום<PDI> hello`. The DOCX test that pinned "AUTO is not passed on" now asserts the
opposite in three cases, including the neutral-prefix one.

Full reactor gate green here and at the tip of the stack; #538 and #539 are rebased onto
this and pushed.

DemchaAV:
Pushed `4efcd7d8`: the duplicated comment in `resolveAlign()` is gone — that was mine,
left behind when I restored the fix after checking the test went red without it — and the
method Javadoc no longer credits `Character` with an answer the shared resolver now gives.

On marking `ParagraphDirection` `@Internal`: it already is, through its package.
`com.demcha.compose.document.layout` carries the annotation at package level, which
`InternalAnnotationCoverageTest` asserts — so the class is covered by the same marker as
`LayoutGraph` and `NodeDefinition` beside it. (`engine.text` and `engine.text.bidi` were
genuinely missing it; that is fixed in #538.)

DemchaAV:
Pushed `ef38eab5`. Three corrections from a fresh five-lens review of the stack.

**Two guards asserted around their subject rather than at it.** The slide test checked that
a right-to-left line is *not* in the frame PowerPoint reflows — true of any name at all,
including a renamed shared frame, which the sibling Latin test would force someone to rename
while leaving this one silently passing. It names the frame the path must produce. The
single-run wrapping test is called "is marked" and asserted the span count and the line text,
both of which a fast path skipping the marking would leave exactly as they are; it asserts
the mark now.

**The changelog credited PowerPoint with a direction mark it never receives.** Only Word is
marked — `w:bidi`, which is the one way it can lay out a line opening on a neutral
character. The PPTX writer emits no direction attribute anywhere; each span is pinned in an
absolute frame at the position the page resolved, so the coordinates carry what a mark would
have said. Corrected to say that per backend.

**And a scope limit worth knowing, which nothing recorded.** Direction stops at the
paragraph: text in a table cell goes through the table's own layout, which has no direction
handling. I rendered both into one page and read the content stream — the paragraph emits
`U+05DD U+05D5 U+05DC U+05E9` (reversed for display, correct), the cell emits
`U+05E9 U+05DC U+05D5 U+05DD` (logical order, backwards on the page). Same string, two
results. It predates this PR rather than being introduced by it, so I have documented it in
the changelog and the recipe rather than folded a table-layout change into this one —
someone would otherwise have found it by staring at a finished table and wondering why only
that part was reversed.

Full reactor gate green here and at the tip of the stack.

PR #538, opened 2026-08-10
feat(engine): join Arabic through its presentation forms

DemchaAV:
Stacked on #536 (→ #534 → #535) — review and merge in that order. Closes the Arabic-joining item of [#140](https://github.com/DemchaAV/GraphCompose/issues/140).

## Why

Arabic letters change shape by position, and a font does that through OpenType `GSUB` — which a PDF content stream never executes: `showText` walks the font's `cmap` and nothing else. With #536 the letters already stood in the right order; they stood unjoined, every one in its isolated form. The presentation forms Amiri was selected for carrying (#535) are the one way to reach the joined shapes in a PDF, and this is the shaping step that maps to them.

## What changed

`ArabicShaper` is a pure text-to-text mapping: a per-letter table of the four positional forms for the base block (U+0621–U+064A + tatweel), the four lam-alef ligatures, and the joining walk. Vowel points and the bidirectional format characters are transparent to the join, as Unicode's joining rules say. Shaping is idempotent, and text with no Arabic letter passes through as the same instance.

It runs **before measurement** — on the logical lines of the plain and markdown paths, at token construction on the inline path — because the forms have their own advance widths and the engine's contract is that what is measured is what is drawn.

Each backend keeps doing only what its medium requires:

| Backend | Receives | Why |
|---|---|---|
| PDF | the forms, then mirroring + reversal at the draw seam | its content stream runs no shaping and applies no bidi rule |
| PPTX | base letters, restored via `toBaseLetters` at its seam | PowerPoint shapes Arabic itself; frozen forms would sit in a file users search and copy from |
| DOCX | untouched logical text | Word was never handed forms at all |

Mirroring (UAX #9 L4: `()[]{}<>«»‹›`) lives at the PDF seam next to the reversal and nowhere else — both office engines apply L4 themselves, and a mirror baked into the span would come out doubled there.

A font that carries the letters but not the forms — the `GSUB`-only families, Scheherazade New among them — now degrades at the glyph seam to unjoined base letters instead of `?`, with one warning per font: the joining is lost, the text is not. A lam-alef ligature decomposes back into its two letters. The handler sanitizes **before** reversing so that decomposition keeps its letter order on the page.

Two deliberate trades, both tested: a vocalized lam-alef (a vowel point between the letters) renders as two joined letters rather than the ligature, because folding the pair would strand the mark and make the slide text differ from what the author wrote — the round-trip is exact instead. And a character-split of an over-long Arabic word keeps mid-word forms at the seam.

Also: `*.pdf binary` in `.gitattributes` — without it, checkout treats a committed README preview as text and rewrites it in the working tree, corrupting the preview and blocking branch switches across a change to it.

## Verification

Full reactor gate green; no snapshot or visual baseline moved.

The proofs read the PDF content stream: every Arabic glyph on the page is a presentation form and none a base letter; a parenthesis before Hebrew is drawn facing what it encloses; the slide text carries base letters with no form leaked into it (`PptxRtlParagraphTest`); degradation is exercised against a mocked `GSUB`-only font, including the two-letter ligature decomposition. Shaper table tests pin every letter's forms, the four ligature variants, transparency of harakat and direction marks, idempotence, and the same-instance fast path. ~30 new tests across core, render-pdf and render-pptx.

The published `text-direction` example preview is regenerated — its Arabic row now shows the joined forms it produces.

Performance: the left-to-right cost is one early-exiting scan per line/token. A first A/B run against the base branch showed sign-flipping deltas with one scenario above the harness's noise band; a clean re-run is in flight and its table will be posted as a comment before this merges.

## Known limits

- Shaping covers the base Arabic block; the Persian and Urdu extensions (پ چ ژ گ …) render unjoined and break joins with covered neighbours. Named in the recipe.
- A mid-word style boundary (bold half-word) shapes each fragment separately, producing word-edge forms at the seam.
- Header/footer, watermark and debug-label text is not shaped.


DemchaAV:
A/B result, as promised in the description. `./scripts/ab-bench.sh -a feat/rtl-bidi-reordering -b feat/rtl-arabic-shaping -r 3 -p smoke` — three interleaved runs per branch, medians compared, all-LTR corpus so this measures what existing documents pay for the shaping scan.

| Scenario | Avg | Docs/s | | First run (pre-fix commit) Avg |
|---|---|---|---|---|
| cv-template | +2.32% | −2.09% | | −1.06% |
| engine-simple | −5.58% | +5.70% | | +7.87% |
| feature-rich | +5.11% | −4.85% | | +0.06% |
| invoice-template | +2.44% | −2.36% | | +10.20% |
| long-token | +0.55% | −0.59% | | −6.73% |
| proposal-template | +2.84% | −2.69% | | +5.93% |
| vector-rich | +0.26% | −0.19% | | +4.24% |

Every average sits inside the harness's declared noise band (~5–10% on a laptop), and the two runs disagree on sign for three scenarios with no code difference on the LTR path between them — the first run's `invoice +10.2%` is `+2.4%` here. That is noise's portrait, not a regression's; the real regression this feature line once had (+10…+17% consistently across every scenario and stage at once) looked nothing like this. Left-to-right cost of shaping is one early-exiting scan per line/token, pinned by the same-instance test.

DemchaAV:
Pushed `355a4695`, which closes the joining-control gap.

**An author can now say where Arabic must not join.** U+200C forbids a join the letters
would otherwise make and U+200D forces one; both are category C, so control sanitizing
deleted them before the shaper — the only thing that reads them — ever ran, and a
document could not express the distinction at all.

The shaper needed almost nothing for the non-joiner: it is an opaque non-joining
character, so the lookups that walk letters already reached the right answer once it
survived the seam. The joiner had to be taught, since it stands in for a letter that is
not there. Both are consumed during shaping — past that point they are code points no
font can encode, measured as one thing and drawn as a substitution mark — and the glyph
seam drops them for text that never reaches the shaper at all.

They get their own predicate rather than joining the bidirectional controls: these say
nothing about direction, only whether two letters connect.

Two other corrections in the same commit:

- Auto-size measured a highlight run unshaped while the token built from it is shaped, so
  a chip was sized against a string the layout never uses. Both paths shape now. I could
  not make a regression test for it go red, and the reason turned out to be a separate,
  larger defect: auto-size does not shrink a paragraph holding a chip **at all**, for any
  script. Measured with the Latin word `Internationalization` at page widths 90–260
  (10pt margins) — plain text goes 7.4 → 11.6 as the column widens while the chip stays
  at 221.99 / 184.99 / 147.99, constant multi-line blocks. That is pre-existing and
  unrelated to Arabic; it is tracked separately rather than folded in here.
- The degraded-shaping warning is keyed on the font rather than on `(font, code point)`.
  This PR's description promised one warning per font; the code gave one per distinct
  letter, so a document turned a single fact about the font into dozens of identical
  warnings.

Tests: four on the shaper (non-joiner breaks a join, joiner forces one, neither survives
into shaped text, neither lets lam-alef ligate), two on the sanitizer for both seams, and
one end-to-end in `RtlGlyphOrderTest` that renders `beh + ZWNJ + heh` and compares the
drawn glyphs against the joined form — the piece no unit test of either half could show,
since each was correct on its own.

Full reactor gate green.

DemchaAV:
Pushed `9d2cf46b`. The PPTX blocker is real and is closed — I reproduced it first: a test
asserting the author's non-joiner reaches PowerPoint was red, exactly as described.

**The fix is not the logical/shaped split, because the reason for consuming the controls
turned out not to hold.** `PdfFont.getTextWidth` sanitizes before measuring, so
measurement and drawing pass through the *same* glyph seam. A control that no font can
encode is therefore removed from both at once and nothing can drift — which means it is
free to stay in the span, where a backend with a shaper of its own can still see it. So
the shaper no longer eats them.

Two seams had to stop removing them. `ParagraphWrapping.directionalLine` ran the sanitizer
that strips direction marks, which I had widened to strip every formatting control; that
is back to the bidirectional ones, since those have been read by the time a line is split
into runs and the joining controls have not. And the slide backend sanitizes with a
variant that keeps them: a PDF drops a zero-width control because it has no glyph to draw
for it, which is not a reason that applies to a backend emitting text.

The naming follows the behaviour, as you suggested — `removeExceptDirectionMarks` keeps
more than direction marks and is now `removeExceptFormattingControls`, while
`removeDirectionMarks` again removes exactly what it says.

**Transparent marks — confirmed and fixed.** `beh + U+0610 + teh` shaped to `FE8F FE95`:
the word came apart where an author added an honorific. Transparency is now decided by
general category, which is Unicode's own rule for anything it does not list explicitly
(Mn, Me, Cf → joining type T), with the joining controls as the listed exception. That is
also shorter than the ranges it replaced. `beh + U+0610 + teh` now shapes to
`FE91 U+0610 FE96`, and the vowel-point case the old ranges did cover is held alongside it.

`engine.text` and `engine.text.bidi` are tagged `@Internal` and `InternalAnnotationCoverageTest`
now reads them. Worth knowing: **no** engine package carried the annotation — this fixes
the two this release touches, not the ~40 others, which is a repo-wide follow-up rather
than something I have quietly finished.

On `PdfFont` in the PPTX backend — that dependency predates this work and is deliberate:
the slide backend is geometry-identical to the PDF one and takes its metrics from there.
The RTL work uses that seam, it did not create it, and decoupling PPTX from PDFBox means
extracting font metrics into a neutral SPI — its own piece of work, not a change to make
inside a shaping PR. The half of that finding that *was* about this PR — core holding only
a shaped representation — is what the fix above addresses.

Full reactor gate green here and at the tip; #539 is rebased onto this.

DemchaAV:
Pushed `1b99cfb9` with the three smaller corrections, and one verification worth reporting.

- `resetForTesting` clears the per-font set for the degraded-shaping warning. It clears the
  state that makes a warning fire once, and the set I added for that warning was not in it,
  so a test expecting it passed or failed on whether something earlier in the same JVM had
  already tripped it.
- The joining-controls paragraph moved above `@param`, where javadoc still reads prose as
  description, and it no longer says the shaper *consumes* them — it stopped doing that when
  they were left in the text for a backend with its own shaper.
- The root pom named the fonts line as `1.0.0`. **I checked this twice in earlier rounds and
  reported it as not present, and I was wrong both times** — the sentence is split across
  lines (`fonts is` / `at its own 1.0.0 line`) and my single-line grep never matched it. It
  now points at the module and the property rather than repeating a number that goes stale
  every font release.

**On breaking a long Arabic word across lines — confirmed.** I rendered forty joined letters
into a narrow column: the word does break, and the seam keeps its connecting strokes (the
line ends on a medial `U+FE92` and the next opens on a final `U+FE90`).

I am not fixing it here, and the reason is structural rather than reluctance. Shaping runs
before wrapping because wrapping measures widths and the contextual forms are what carry
them — an isolated form is wider than a medial one. Re-shaping the two halves after the
break changes their widths, which is exactly what the wrap has already spent, so it needs
re-wrapping, which can move the break, which re-shapes again. That is a real algorithmic
change, not a patch. It also only arises where a single word cannot fit its column at all —
the case every script degrades in, and one Arabic typography avoids by not breaking words.
The changelog now states the limit rather than leaving it to be found.

**On the architecture.** The two reviews you have differ here: the other one reads the same
code and concludes it would not introduce `logicalText + renderText` for a single shaper,
and that the transformation is effectively reversible now that base letters, marks and
joining controls all survive. I agree with that reading, and I would add that the `PdfFont`
dependency in the slide backend predates this work — the backend was built geometry-identical
to the PDF one and takes its metrics from there. Decoupling it means extracting font metrics
behind a neutral SPI, which is its own piece of work with its own risk, and folding it into
a shaping PR would make both harder to review. The half of that finding that *was* about
this PR — core keeping only a shaped representation, so an author's instruction could not
survive to PowerPoint — is fixed and has an end-to-end test.

Full reactor gate green.

DemchaAV:
Pushed `25d7aee8` with three corrections from a fresh review of the stack.

**A sentence that contradicted itself.** The changelog said "a mathematical operator passes
through drawn as written" — in a sentence that had already listed angle brackets among what
*does* mirror. `U+003C` and `U+003E` are both: Unicode classes them as mathematical, and
they are in the table. It now names a relational or set operator (`≤`, `⊂`) instead, so the
same character no longer sits on both sides of the claim.

**A hot-path walk for a certain miss.** `baseLettersOf` walked the ligature list and all
forty-two letter entries before concluding a code point is not a presentation form, and its
caller on the glyph seam asks exactly that about every code point a font cannot encode, on
each occurrence — so a Latin document with one unsupported bullet paid the walk per bullet.
Everything the shaper emits lives in one block, so a bounds check answers the miss.

**And the escapes comment was wrong again.** For the third time in this work a comment
claimed ZWNJ/ZWJ were written as escapes while the literals held the raw invisible
characters. Fixed, and the recurrence is the useful signal: the way to stop it is to build
these from code points (`String.valueOf((char) 0x200C)`) so there is nothing invisible to
get wrong, which is what the isolate constants in `ParagraphDirectionTest` already do.

For the record on the wider review: 29 findings were raised across five lenses, 7 survived
adversarial refutation. The refuted set included the concern that this stack regresses
plain-LTR documents, and that PPTX now forwards the joining controls for non-Arabic scripts
— the latter is intended, since PowerPoint's own shaper is the reader those characters were
written for.

Full reactor gate green.

PR #540, opened 2026-08-11
refactor(fonts): let the font artifact state its own version

DemchaAV:
## Why

Telling a *stale* font artifact apart from a *missing* one needed the engine to know
which release introduced which family. That was a map from resource folder to version —
five entries by the time the Korean font landed, one more for every font after it —
living in `FontFamilyDefinition`, a class that otherwise knows nothing about which fonts
exist. Adding a family meant remembering to edit it, and forgetting meant falling back to
advice that does not apply: *add a dependency* to a reader who already has it.

Moving that map into a catalog descriptor would relocate the coupling rather than remove
it — the catalog would then be carrying the artifact's release history, which it does not
carry today.

## What

The artifact answers for itself. The `fonts` module writes a small descriptor into its
jar naming its own version, and the engine reads it:

- no descriptor → the artifact is not on the classpath → the message points at the
  v1.8.0 migration note, which is the advice that fits;
- a descriptor → the artifact is present and older than what was asked for → the message
  names **the version the consumer actually has**.

That is the more useful answer as well as the cheaper one. *"The graph-compose-fonts on
your classpath is 1.0.0 and does not carry this"* identifies the problem from the
reader's side; *"this family needs 1.1.0 or newer"* asserts something the engine can only
know by being told — and told again for every font after it.

Gone with the map: the `Lato-Regular.ttf` probe (artifact presence was inferred from one
long-standing face, which happened to work rather than meaning anything) and the path
parsing that extracted a family folder.

**Resource filtering is scoped to the descriptor alone.** Filtering rewrites the bytes of
every file it touches and a font binary does not survive it, so the fonts stay in an
unfiltered resource block that excludes the one text file. Verified against the built jar:
the descriptor carries `version=1.1.0`, `GothicA1-Regular.ttf` is byte-identical to its
source, and all 130 faces are present.

## Tests

- The five bundled families that need the diagnosis are driven through it, asserting the
  version read from the artifact's own descriptor rather than a literal — so this stays
  true across a version bump with nothing to edit.
- Both branches of the message are now tested directly. The absent-artifact branch was
  untestable before, because the artifact is always on the test classpath, which left the
  wording for the one setup nobody here is in free to rot; the builder takes the version
  explicitly so both can be driven.
- A missing face of a long-standing family is diagnosed like any other. That test used to
  assert it received the *add the dependency* advice — which was wrong, since the
  dependency is there.
- Full reactor gate green (`clean verify` across core, render-pdf/docx/pptx, templates,
  testing, qa, coverage).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -143,7 +143,10 @@ follow semantic versioning; release dates are ISO 8601.
   consumer actually has.
 
   The artifact states that version itself: it ships a small descriptor written at build
-  time, and the engine reads it. The first shape of this carried a map from family folder
+  time, and the engine reads it. Presence is asked separately, of a face the artifact has
+  carried since its first release — the descriptor only ships from 1.1.0, so for the
+  releases published before it a missing descriptor means "too old to say", not "not
+  here", and those two need opposite advice. The first shape of this carried a map from family folder
   to the release that introduced it, which meant the catalog had to remember the
   artifact's history and every new font needed an entry in a class that otherwise knows
   nothing about which fonts exist. Reading the version off the artifact answers the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -139,8 +139,16 @@ follow semantic versioning; release dates are ISO 8601.
 - **A stale font artifact is now distinguished from a missing one.** Asking for a
   family that a newer `graph-compose-fonts` introduced produced the same message as
   having no font artifact at all — an instruction to add a dependency that was already
-  there. The two causes are now told apart, and the stale case names the version that
-  carries the family.
+  there. The two causes are now told apart, and the stale case names the version the
+  consumer actually has.
+
+  The artifact states that version itself: it ships a small descriptor written at build
+  time, and the engine reads it. The first shape of this carried a map from family folder
+  to the release that introduced it, which meant the catalog had to remember the
+  artifact's history and every new font needed an entry in a class that otherwise knows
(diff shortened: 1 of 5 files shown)


DemchaAV:
**A fresh review of the stack found a regression this PR introduced, and it is fixed in
`be1d08fe`.** Reporting it plainly because the original commit message argued the opposite.

Reading the diagnosis off the version descriptor alone broke it for the only people who can
reach it. The descriptor ships from **1.1.0**; the one release published to Central is
**1.0.0** (verified: the 1.0.0 POM returns 200). So a consumer holding the published fonts
artifact has the fonts and no descriptor — which the code read as *no artifact on the
classpath* and answered with "add the dependency graph-compose-fonts". Advice for a build
that lacks it, given to the build that has it: precisely the wrong answer this whole
diagnosis exists to stop giving.

Presence and version turn out to be two questions, and only one of them can be asked of a
release that predates both mechanisms:

- the descriptor answers **which version**, for releases new enough to carry it;
- a face the artifact has carried since its first release answers **whether it is there at
  all**, for any version ever published.

So the long-standing-face probe is back, and the commit that removed it was wrong about it.
It did not merely happen to work — it is the only signal that survives an artifact older
than anything we can ship inside one. A missing descriptor with the artifact present now
reads as "predates 1.1.0" rather than "absent". Three states, three tests, including the one
describing every consumer of the currently published release.

Full reactor gate green at this tip.

PR #555, opened 2026-08-14
perf(docx): resolve an exported image once instead of twice

DemchaAV:
Closes the remaining half of #531.

## Why

`DocxSemanticBackend.writeImage` needs the node's data twice over — the bytes it writes and the intrinsic size it measures the fit against, then the box the frame gets — and asked for it twice:

```java
ImageData resolved = NodeDefinitionSupport.toImageData(node.imageData());   // once
…
NodeDefinitionSupport.resolveImageDimensions(node, contentWidth);           // and again, inside
```

Resolving is not free. `ImageSourceCache.fromBytes` copies the byte array whole (`Arrays.copyOf`) and takes a SHA-256 over it, so a 10 MB image paid two full copies and two hashes on every export. A path-sourced image also announced its arrival twice in the log, which is what made the duplication visible in the first place.

The other half of #531 — a row cell dropping its paragraph alignment — was fixed in [#547](https://github.com/DemchaAV/GraphCompose/pull/547), where the cell walk started going through `applyParagraphProperties`.

## What changed

`resolveImageDimensions` gains a three-argument overload taking the `ImageData` the caller already holds. The existing two-argument form resolves and delegates to it, so the layout path (`ImageDefinition`) is untouched and the DOCX backend is the only caller that changes.

## Verification

`DocxImageResolutionTest` exports a path-sourced image and counts the arrival messages: one image, one resolution. Red before the fix, with the message logged twice —

```
messages were [Create an image from path: …probe.png, Create an image from path: …probe.png]
```

Counted through the log rather than through a spy on purpose: `DocumentImageData` is `final`, so there is no seam to instrument, and instrumenting production code to make a cost measurable would put test-only machinery in `src/main`. The arrival message is the one observable the resolution already leaves behind.

What the test cannot see directly is the copy and the hash — those have no observable of their own. It counts the resolution, and the copy and hash are what a resolution costs.

`./mvnw clean verify` over the full reactor — 13/13 modules, `BUILD SUCCESS`.

### DIFF
--- a/core/src/main/java/com/demcha/compose/document/layout/NodeDefinitionSupport.java
+++ b/core/src/main/java/com/demcha/compose/document/layout/NodeDefinitionSupport.java
@@ -293,7 +293,26 @@ public final class NodeDefinitionSupport {
      * @return resolved image dimensions excluding padding
      */
     public static ImageDimensions resolveImageDimensions(ImageNode node, double availableWidth) {
-        ImageData imageData = toImageData(node.imageData());
+        return resolveImageDimensions(node, availableWidth, toImageData(node.imageData()));
+    }
+
+    /**
+     * As {@link #resolveImageDimensions(ImageNode, double)}, for a caller that already
+     * holds the node's resolved data.
+     *
+     * <p>Resolving it is not free: {@code ImageSourceCache.fromBytes} copies the byte
+     * array and hashes it whole, and a path-sourced image logs its arrival. A caller that
+     * needs the data <em>and</em> the dimensions — the DOCX backend writes the bytes and
+     * sizes the frame — paid all of that twice for one image, which on a large one is two
+     * copies and two SHA-256 passes.</p>
+     *
+     * @param node           image node to measure
+     * @param availableWidth parent available width
+     * @param imageData      the node's already-resolved data; must be the node's own
+     * @return resolved image dimensions excluding padding
+     */
+    public static ImageDimensions resolveImageDimensions(ImageNode node, double availableWidth,
+                                                         ImageData imageData) {
         int intrinsicWidth = Math.max(1, imageData.getMetadata().width());
         int intrinsicHeight = Math.max(1, imageData.getMetadata().height());
         double ratio = intrinsicWidth / (double) intrinsicHeight;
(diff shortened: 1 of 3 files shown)


DemchaAV:
Confirmed, and it is worse than a second read — thank you for pushing on it. I fixed the second *resolve* and left the larger half standing.

**The divergence is real and silent.** `ImageSourceCache.fromPath` keys on the absolute path alone, with no mtime in the key. So once a first render warms it, a file rewritten underneath gives `writeImage` fresh bytes from `readAllBytes` and the previous version's metadata from the cache. Both halves succeed; they simply describe different files, and the document carries a picture embedded at another image's dimensions with nothing reporting it.

Fixed as you suggested — bytes come from the `ImageData` that sized the frame:

```java
ImageData resolved = NodeDefinitionSupport.toImageData(node.imageData());
byte[] bytes = resolved.getBytes();
```

`readBytes` and the `DocumentImageData` import are gone with it.

**On the silent skip:** it was not a contract, it was a side effect — `readBytes` swallowed its exception and returned an empty array, which the length check then treated as "nothing to draw". Resolving throws instead, so the catch is now explicit. I kept it non-fatal, since one unreadable image should not cost the whole document, but it logs the drop the way every other dropped node in this backend already does rather than vanishing.

**The new case pins the divergence, not the cost**, which is the gap you identified in the first test: render, rewrite the file with a differently-sized image, render again, then read the embedded picture back out of the `.docx`. Before the fix it reports

```
expected: 40
 but was: 90
```

— a 90-wide picture in a frame built for a 40-wide one. The arrival-count test stays, since the two failures are different: one is a repeated resolution, the other is two sources disagreeing.

Full reactor `clean verify` — 13/13, `BUILD SUCCESS`.

DemchaAV:
Right on both counts, and the second one is the part I got wrong: I read the old behaviour as "this backend skips unreadable images" and preserved it, when it was a side effect of `readBytes` swallowing its own exception. Preserving it around the *resolver* then widened it enormously — corrupt bytes, a missing `ImageReader`, a metadata decode that gives up and any defect in the cache all became one silently missing picture, announced once per export and only for the first.

**The catch is gone.** A resolution failure reaches the caller, which is exactly what it did for every one of those cases before this branch touched the method.

The one case that changes is the unreadable path, and it changes for the better — a document that quietly comes back one picture short is the worst of the available answers. I pinned it with `anImageThatCannotBeReadFailsTheExportRatherThanVanishing`, so it reads as a decision rather than as a side effect of deleting code.

I have not introduced a typed source-read exception here. Agreed that it belongs in its own change with one policy across the backends, rather than being invented in a PR about resolving an image once.

Full reactor `clean verify` — 13/13, `BUILD SUCCESS`. Three cases in the suite now: one resolution per export, bytes and size from that same resolution, and an unreadable source failing loudly.

PR #556, opened 2026-08-14
docs: give coding agents the rules this repository already runs on

DemchaAV:
## Why

An agent cloning this repository is told nothing about it. `CONTRIBUTING.md` and the architecture docs carry the lanes, the pipeline and the testing expectations, but the operating rules — what not to touch, which results lie — lived in whatever local configuration a maintainer happened to have. Nothing in the tree.

Three build properties are worse than undocumented, because each returns success while reporting the opposite of the truth:

- **A standalone examples run resolves from `~/.m2`.** The reactor builds `examples` from source; `cd examples && ../mvnw compile exec:java …` does not — exactly as `CONTRIBUTING.md:37` already warns for `qa`. Skip the install after a render change and the examples regenerate through the artifacts you last installed, so the output looks unchanged because the change was never in it.
- **Example output is not byte-comparable.** The writers embed timestamps, so a fresh render differs from its committed preview for *every* example, including ones the change cannot reach. Verifying by `cmp` concludes the opposite of the truth.
- **The tree collects zero-byte files** named by shell quoting accidents (`$env`, `` b)` ``, `'Current`). `git add .` commits them; `git clean -f` removes new sources along with them.

## What's here

**`AGENTS.md`**, 118 lines. It links `CONTRIBUTING.md` and the architecture docs and answers only what none of them does: release work is not a side effect of a feature branch, staging is by path, the two misleading results above, and "never report a test passed unless it was executed".

An earlier revision of this branch was 440 lines and restated branch policy, the Java baseline, the pipeline, package ownership, module responsibilities, the API tiers and the POM rules — while opening by calling those documents the source of truth. That is a second copy that goes stale the day the original moves, and this repository has already lost a full release line to exactly that: prose describing an engine it no longer has, green under every guard. Determinism was checked too and delegated, since `overview.md` and `package-map.md` each cover it.

**`AgentsGuideGuardTest`**, three cases, holding only what it can see:

| Case | Red when |
|---|---|
| `everyPathItSendsAReaderToExists` | a linked document has moved |
| `theBuildCommandsItPrescribesCanBeRun` | the reactor gate is dropped from the file, or `mvnw` is not at the root |
| `itDoesNotRestateThePackageCoordinatesItLinksTo` | a `com.demcha.compose.*` coordinate reappears |

The third guards the regression above: existence checks cannot see a restatement go stale, because the packages still exist — they have simply stopped being the ones the guide names.

**`.gitignore`** — `/AGENTS.md` was ignored, listed among scratch files (`rectangle.pdf`, `Test.pdf`), from when it was one. Removed, or none of this could ship. `/CLAUDE.md` added: a file named for one tool is configuration belonging to whoever runs that tool, and committing one would oblige this repository to carry a copy of these rules per vendor — the duplication `AGENTS.md` exists to avoid.

**`.github/workflows/ci.yml`** — the guard joins the `Architecture and Documentation Guards` list, which was not running it.

## Verification

`./mvnw -B -ntp clean verify` over the full reactor — `BUILD SUCCESS`. The 9-guard fail-fast list runs 50 tests green.

Both commands the file prescribes were run before it prescribed them: `./mvnw -B -ntp verify -pl :graph-compose-core` and the bare `./mvnw -B -ntp clean verify`. Every document it links was checked to open.

Each guard case was checked against its own breakage — a moved link, a dropped gate command, a package coordinate added back: 2 of 3 red in every case.

## Scope

These are the maintainer's existing practices, not new policy, and nothing here changes how the build works.

Two things were deliberately left out. A commit-attribution rule from an earlier draft governs the maintainer's own commits on their own machine and is not project policy — `CONTRIBUTING.md` has no DCO, sign-off or co-authorship requirement — so it does not belong in a file addressed to contributors. And `PackageMapGuardTest`'s reactor-module logic was not reused: the module inventory it would have checked is precisely the restatement this branch removed, and a stronger check on a paragraph that should not exist is still a paragraph that should not exist.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,7 @@ jobs:
         # in qa) run in build-and-test below, which now also covers docs-only PRs.
         run: |
           ./mvnw -B -ntp clean \
-            "-Dtest=EnginePdfBoundaryTest,DocumentationCoverageTest,CanonicalSurfaceGuardTest,PackageMapGuardTest,VersionConsistencyGuardTest,CiGuardListGuardTest,CiGateCoverageGuardTest,CodeQlScopeGuardTest" \
+            "-Dtest=EnginePdfBoundaryTest,DocumentationCoverageTest,CanonicalSurfaceGuardTest,PackageMapGuardTest,VersionConsistencyGuardTest,CiGuardListGuardTest,CiGateCoverageGuardTest,CodeQlScopeGuardTest,AgentsGuideGuardTest" \
             test -pl :graph-compose-core
 
   changes:
--- a/.gitignore
+++ b/.gitignore
@@ -52,7 +52,6 @@ render-pdf/logs/
 /patch-root-position.diff
 /rectangle.pdf
 /Test.pdf
-/AGENTS.md
 
 ### Editor / OS scratch files ###
 *.tmp
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,11 @@ render-pdf/logs/
 /patch-root-position.diff
 /rectangle.pdf
 /Test.pdf
+# Agent config that is not the project's to publish. AGENTS.md is the generic,
+# tool-independent guide and ships; a tool-specific file stays on the machine
+# that uses it, so the repository carries one set of rules rather than one per
+# vendor.
+/CLAUDE.md
 
 ### Editor / OS scratch files ###
 *.tmp
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,440 +1,118 @@
 # AGENTS.md
 
-## Purpose
+Operating rules for coding agents working on GraphCompose.
 
-This file defines the operating rules for coding agents working on GraphCompose.
+GraphCompose is a modular Java document layout engine, not a generic Java
+application. Its architecture, module boundaries, public API tiers and
+deterministic layout behaviour are documented — read them before changing them.
+This file does not restate them. It covers what an agent needs that those
+documents do not say, and points at the rest.
 
-GraphCompose is a modular Java document layout engine.
-Changes must preserve its architecture, module boundaries, public API stability,
-deterministic layout behaviour, and backend independence.
+## Read first
 
-Do not treat this repository as a generic Java application.
+- [`CONTRIBUTING.md`](CONTRIBUTING.md) — contribution flow, branch policy, Java
+  baseline, repository map, architecture lanes, testing expectations
+- [`docs/architecture/overview.md`](docs/architecture/overview.md) — the
+  authoring pipeline and where each stage's responsibility ends
+- [`docs/architecture/package-map.md`](docs/architecture/package-map.md) —
+  package ownership and module layout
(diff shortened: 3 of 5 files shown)


DemchaAV:
All three fixed, and Major 2 was the one worth stopping for — the file opened by calling the architecture docs the source of truth and then restated them. I wrote the contradiction into the first paragraph and did not see it.

**Major 2 — 440 lines to 118.** Branch policy, Java baseline, pipeline, package roots, module responsibilities, API tiers, POM rules: all of it lives in `CONTRIBUTING.md`, `docs/architecture/overview.md`, `package-map.md` and `docs/api-stability.md`, and all of it is now a link. I checked determinism too, since it was the one section I thought might be homeless — `overview.md` and `package-map.md` cover it five times each, so it went as well.

What is left is what none of them says: release work is not a side effect of a feature branch; stage by path because the tree collects shell-quoting junk; and the two commands that return success while reporting the opposite of the truth — a standalone examples run resolving from `~/.m2`, and example output that is never byte-comparable.

**Major 1 — the false promise is gone with its subject.** You were right that `everyModuleItNamesIsBuilt` read the module inventory and then asserted only that `<module>/pom.xml` existed, so dropping `<module>qa</module>` from the reactor POM left it green while the comment claimed otherwise.

I did not reuse `PackageMapGuardTest`'s `REACTOR_MODULE` in the end, because the inventory it would check is exactly the restatement Major 2 asks me to delete — a stronger check on a paragraph that should not exist is still a paragraph that should not exist. The manual `SOURCE_ROOTS` went with it.

The guard now holds three things it can actually see:

| Case | Red when |
|---|---|
| `everyPathItSendsAReaderToExists` | a link target is moved |
| `theBuildCommandsItPrescribesCanBeRun` | the reactor gate command is dropped, or `mvnw` is not at the root |
| `itDoesNotAnswerWhatItDelegates` | a `com.demcha.compose.*` coordinate reappears |

The third is the one that guards your Major 2 against coming back: existence checks cannot see a restatement go stale, because the packages still exist — they have simply stopped being the ones the guide names.

Each verified against its own breakage: 2 of 3 red in every case.

**Minor** — added to the `Architecture and Documentation Guards` job. `CiGuardListGuardTest` requires every name in that list to resolve under `core/src/test`, and the 9-guard list now runs 50 tests green.

Full reactor `clean verify` — `BUILD SUCCESS`.

DemchaAV:
Both fixed in `49e61297`.

**The name.** `itDoesNotAnswerWhatItDelegates` claimed the whole delegation contract while checking one recognisable half of it. Whether a paragraph has quietly become a copy of the branch policy or the module roles is not something a regex can tell from prose — and a name implying otherwise is the same over-promise the reactor check made before it was removed. It is `itDoesNotRestateThePackageCoordinatesItLinksTo` now, and the class javadoc says the same.

**The PR body** is rewritten for what is actually here. Worth doing beyond tidiness: with squash-merge this body becomes the commit message on `develop`, so a description of a 440-line file that no longer exists would outlive the file itself — the exact shape of drift this branch is about.

Full reactor `clean verify` — `BUILD SUCCESS`.

PR #630, opened 2026-08-31
chore(knowledge): generate the public API surface from the classes and gate it

DemchaAV:
## Summary

GraphCompose's public API was described authoritatively **outside** GraphCompose. Agents authored against `.llm-wiki/00-api-surface.md` — a gitignored file, generated by a regex parser, quoted to them as a closed set ("if a method is not listed here, it does not exist"). Being gitignored, it was invisible to CI, and it drifted through the whole 2.0 major with nothing watching.

It was also wrong in a way that reads as authoritative. Sections are emitted per *file*, so every nested public type loses its receiver: `### GraphCompose` lists 25 methods of which **7 are real** — the other 18 belong to the nested `GraphCompose.DocumentBuilder`, which has no section at all. It omits `graph-compose-testing`, `-render-docx` and `-render-pptx` entirely, and publishes `document.dsl.internal` as callable.

This replaces it with a tracked `knowledge/` pack generated from the compiled classes and gated in CI, plus the three layers that make it useful: what each page promises, which way to do a thing, and which pages a change invalidated.

Five commits, reviewable in order.

## Commits

1. **`ea3a3ff5` — the surfaces, and the gate.** `knowledge/tools/api-surface/extract-api.mjs` reads `*/target/classes` with `javap` (`--from-reactor`) or a published release through Maven (`--from-release`), and writes one document per surface: `authoring` (227 types), `templates` (161), `backends` (69), `extension-spi` (9), `testing` (5).

   Classification runs in three stages, because admission and stability are different questions and `@Beta` answers only the second: a type enters `extension-spi` by an authored SPI list, never by carrying `@Beta`. `docs/api-stability.md:33-34` gives Extension SPI and Experimental the *same* annotation and puts the distinction in the docstring, so no classifier can route between those surfaces on its own.

   Annotations are read from the class file, not the source. Both spellings occur here — `svg/SvgIcon.java` writes `@Beta`, `layout/package-info.java` writes `@com.demcha.compose.document.api.Internal` — and a grep for either form returns the wrong set. Members are keyed by erased parameter types rather than arity: `ShapeContainerBuilder` declares two `path` overloads with three parameters each and only the `SvgPath` one is `@Beta`.

   A public type matching no rule **fails the run**. That is what surfaced `engine.components`, which despite its name holds `TextStyle`, `DocumentMetadata`, `ImageData`, `Padding` and `Anchor` — value types the public API makes you construct, and absent from the old hand-written package list. They are admitted because admitted signatures mention them; their engine-side twins (`HeaderFooterConfig` beside the public `DocumentHeaderFooter`) are mentioned by nothing and are not. Every exclusion is written to `knowledge/api/excluded.json` with its reason.

   `knowledge/manifest.json` carries only stable facts. Commit, timestamp and artifact digests go to `target/knowledge/provenance.json` and are never committed: a tracked file cannot hold the SHA of the commit it belongs to, and class digests are not reproducible across machines, so a gate comparing them would be red on every run.

2. **`f86b1f2d` — claims.** A page states what it promises, in a marker beside the prose that makes it:

   ```markdown
   <!-- claim: symbol=TableBuilder.zebra -->
   <!-- claim: behavior=table.explicit-row-style-beats-zebra proof=test:TableBuilderZebraAndTotalsTest -->
   ```

   Front-matter was not an option: none of the 80 tracked public pages carries any and there is no Jekyll config, so `docs/` renders straight from GitHub, where it would appear as a rule and literal `key: value` at the top of every page. An external file keyed by path + anchor breaks silently when a heading is edited. A marker is invisible, moves with the paragraph, and is the convention `doc-example` already set.

   Enforcement is asymmetric. A claimed symbol no surface has **fails** — the page describes API a reader will try to compile. A symbol an example calls without claiming is reported only; demanding a claim for every incidental type is churn people learn to skip. `test:` and `snippet:` proofs are resolved, so a proof orphaned by a rename fails rather than quietly holding nothing up, and a `behavior` claim with no proof is refused outright.

3. **`445ba508` — routing.** The surfaces say what exists; they cannot say that a skills list in two columns is a row with weights and not a table.

   ```
   $ node knowledge/tools/api-query/api-query.mjs --task layout.two-columns
   use: row-with-weights
   instead, when: table — the columns are a data grid with a header row
                  costs: paginates per row, sizes by table rules not weights
   constraints: row.rejects-a-nested-row
   read: docs/recipes/layered-page-design.md#sidebar-page-background-vs-row
   ```

   Alternatives are objects, never bare names: a name says a second way exists without saying when it wins. A route is served only once its anchor resolves to a heading that exists, every symbol is in a surface, and every constraint names a documented behaviour some test holds up. Authoring the first route caught a symbol that does not exist — the canvas builder is `CanvasLayerBuilder`, not `CanvasBuilder`.

4. **`641bf490` — the bundle.** A tracked directory is not a published pack; without an archive, "the plugin consumes GraphCompose's knowledge" means "the plugin needs a checkout". `graph-compose-knowledge-<version>.zip` carries the surfaces, routing, claims, provenance — and `bin/query.mjs`, a query CLI of its own, because the acceptance test is that it answers on a machine with no GraphCompose source and there would otherwise be nothing there to run the test with.

   `<bundle>.zip.sha256` is published *beside* the archive: a hash stored in the file it describes is rewritten by whatever rewrote the file. `bundle-checksums.json` inside then says which entry is wrong. Zip timestamps are fixed, so two builds of identical content produce an identical archive.

   Distribution is a GitHub Release asset, not a Maven artifact: the consumer resolves nothing through Maven and no Java build compiles against a documentation pack.

5. **`77ffd5f7` — the changeset.** Both sides of an API comparison are now tracked, so `git show <ref>:knowledge/api/<surface>.json` gives the base and the diff costs no build at all.

   Behavioural signals are ranked: a changed layout snapshot or visual baseline is **proof** — the file *is* the recorded output. A changed example is a **hint**. "A test file changed" is not a signal: tests move for stylistic reasons and a channel that fires on every refactor is one people stop reading.

## Testing

`./mvnw -B -ntp clean verify` over the reactor slice (the ten modules `ci.yml` builds) →
**BUILD SUCCESS** in 2:28, **2098 tests across 375 classes, 0 failures, 0 errors**.

Documentation and CI guards run separately, since this touches `AGENTS.md` and `ci.yml`: `EnginePdfBoundaryTest, DocumentationCoverageTest, CanonicalSurfaceGuardTest, PackageMapGuardTest, VersionConsistencyGuardTest, CiGuardListGuardTest, CiGateCoverageGuardTest, CodeQlScopeGuardTest, AgentsGuideGuardTest, BenchmarkDependencyInstallGuardTest` → **52 tests, 0 failures**.

New checks, all plain Node with no test framework — the repository ships no dependencies and `zip.mjs` was hand-written rather than add one:

- `classifier.test.mjs` — **18 cases**. Compiles Java fixtures during the test against the real `@Internal`/`@Beta` from `core/target/classes`, and covers each rung: unruled type → unclassified; package-`@Internal` excludes; a `@Beta` type in an `@Internal` package is *not* admitted by its `@Beta`; type-level and nested-type stability; a nested type not infected by its enclosing type; member `@Internal` dropped; member `@Beta`; the same-name-same-arity overload pair discriminated.
- `claims.test.mjs` — **18 cases**, mostly negative, including that a nested receiver resolves to the nested type rather than crediting its outer.

Each suite was mutation-tested rather than watched passing. Disabling package-`@Internal` exclusion turns two cases red; disabling the member drop turns one red; reverting the member key to name/arity turns three red, including the overload discrimination that motivated it; reverting the resolver to split on the first dot turns the nested-receiver case red.

The gate itself was seen failing end to end: a throwaway public class compiled straight into `core/target/classes` made `--from-reactor --check` exit 1, name `knowledge/api/authoring.json` and `authoring.md`, and print the regenerate command; removing the class returned it to 0.

The changeset pipeline was verified by replaying a real removal — `@Internal` on both `TableBuilder.zebra` overloads, recompiled and regenerated — and it named one page and one heading with no false positives:

```
API: +0 -2 ~0
  - TableBuilder.zebra(DocumentColor,DocumentColor)
  - TableBuilder.zebra(DocumentTableStyle,DocumentTableStyle)
pages to review (1):
  docs/recipes/tables.md  (Zebra — alternating row fills)
```

The claims checker failed independently on the same state, without sharing a code path. Both the source edit and the recompile were reverted.

`build-bundle.mjs --verify` unpacks into a scratch directory away from the repository and queries it there: a nested receiver resolves, an intent resolves, and an invented method exits 3.

## Where the gate runs, and why there

Three steps inside `build-and-test`, guarded `if: matrix.java == '17'`, between the Maven build and Javadoc. Every CI job gets a fresh runner, so a standalone job would have no `target/classes` to read; the JDK guard keeps a JDK-independent surface from being computed three times.

`knowledge/**` joins the `code` path filter — without it a PR touching only the extractor or a generated surface matches no filter, skips `build-and-test`, and skips the gate with it, so the gate would be absent on exactly the PRs that change it. It is deliberately **not** in `jvm`, so such a PR runs one JDK 17 job rather than the full matrix. `ci-gate` needed no change and no new job was added, so `CiGateCoverageGuardTest` is untouched.

## Notes for review

- **No public Java API changes.** The only edits outside `knowledge/` and the workflows are `AGENTS.md` and two `docs/recipes/` pages, which gain invisible HTML comments. `DocumentationSnippetCompileTest` stays green with the markers present. No CHANGELOG entry, on that basis.
- **`tasks.json`, not `tasks.yaml`.** The rationale this schema carries lives in `useWhen`/`tradeoffs` fields rather than comments, so YAML buys nothing that would justify hand-rolling a parser here.
- **Three routing recommendations are unconfirmed.** Every mechanical gate passes, but whether a recommendation is the *right* one is a maintainer's call. All three carry `confirmedBy: null`, and both the checker and the CLI say so in their own output rather than presenting an unreviewed opinion as settled. These need your sign-off: `layout.two-columns` → `row-with-weights`, `layout.choose-the-layer` → `page-flow`, `table.header-on-every-page` → `repeat-header`.
- **A documentation defect is left unfixed on purpose.** `docs/recipes/tables.md` tells readers `TableResolvedCell.yOffset` "is negative for spanning cells". `TableResolvedCell` is engine-side and in no surface, so a reader cannot observe it. Seeding it as a claim would commit a known-failing marker; it belongs in a separate docs fix.
- **`.llm-wiki/tools/api-index/api-index.py` still exists.** It is removed once no consumer reads it; `~/.claude/CLAUDE.md` and both skill trees already carry a warning that its output is not authoritative.
- **`document/svg/package-info.java` claims beta in prose but carries no `@Beta`.** So an unannotated type added to that package would classify `stable` against the package's stated intent. That is a source fix, not a classifier one — a prose-scraping rule would make the classifier's input exactly as untrustworthy as the file this pack replaces.
- **Claim and proof coverage is thin by design at this point.** 25 claims on 2 pages, and 4 of 80 tracked pages carry `doc-example` markers against 50 pages holding 215 fenced Java blocks. The suggestion channel now makes the gap visible per page; widening it is documentation work.
- **`probe:` and `render:` proofs parse but do not yet resolve** — their registries depend on repairing the snippet-smoke and render-proof harnesses, which the module split broke.

**Lane:** test — new guards, generated artefacts and CI wiring; no engine, canonical API or render output touched.


DemchaAV:
All three findings are fixed in `1965e28d`, and the second one led to a fourth that was not in the review.

## 1. Constructor annotations — fixed

Confirmed exactly as described: `javap.mjs:208` renames a constructor to the simple type name while `annotations.mjs` keys it `<init>`, so the two spellings can never meet.

Checked whether anything shipped is affected: **nothing is.** The only annotated constructors in 2.2.2 are two `@Deprecated` ones, which the classifier does not act on. So no surface on disk was wrong — the contract allowed it, and that was enough to fix.

The fix is not a branch at the call site. The key rule now lives in `memberKeyForMember`, which the extractor and the tests both go through, because deriving the key at each call site is precisely how it came to be missing at the only site that mattered. Fixtures cover `@Internal`, `@Beta` and unannotated constructors, and two assertions state the trap directly: javap's spelling finds nothing, and the class file says `<init>(int)`.

## 2. Fail-open annotation reads — fixed

Absent stays fine — most packages have no `package-info` at all. Present-but-unreadable now stops the run with the class or package named. The reader throws on an unknown constant-pool tag *because* it cannot know what it is looking at, and swallowing that turned "I could not tell" into "there is nothing here".

## 4. A class javap could not read simply vanished — found while testing 2

Corrupting a class to exercise the new fail-closed path did not trigger it. javap exited **1**, the extractor exited **0**, and `SvgIcon` was gone from the surface with no complaint — javap runs in batches, and a bad class does not stop the batch: it reports on stderr and returns the rest.

That is the same shape as the drift this pack exists to end, arriving from inside its own tooling, and it sat one layer below where the review was looking. Every candidate that goes into javap must now come back out; a corrupt class fails the run and names itself.

## 3. Release integrity — fixed

`check-claims.mjs --check` and `check-routes.mjs` now run before `build-bundle`. The bundle carries `claims/` and `routing/`, so a correct API surface was never enough to publish it.

One deviation from the suggestion: `check-routes` is called **without** `--check`. The flag was parsed and never read, so it advertised a mode the tool does not have. Routes are hand-authored with no generated counterpart, so validating them *is* the check; the flag is removed rather than left as a no-op.

## Verification

Reactor slice `clean verify` → **BUILD SUCCESS, 2098 tests across 375 classes, 0 failures, 0 errors**.

Knowledge gates: `classifier.test` **23 passed** (up from 18), `claims.test` 18, `extract-api --check` current, `check-claims --check` current, `check-routes` 3 routes, `build-bundle --verify` answers standalone.

Both new fail-closed paths were verified by causing them, then reverted: a corrupt class file makes the run exit non-zero naming `com.demcha.compose.document.svg.SvgIcon`, and the tree returns to green afterwards.

**No generated surface changed.** `knowledge/api/*` differ only in line endings — the constructor fix alters no output today, which is what latent means.

DemchaAV:
Rebased onto `chore/open-2.3.0` and regenerated the pack against 2.3.0-SNAPSHOT (`2d5694ae`).

**The base change made the gate fire, and it was right to.** Retargeting moved the base past `chore(release): open 2.3.0-SNAPSHOT`, so the committed pack still described 2.2.3-SNAPSHOT. Every check on this PR stayed green through that, because GitHub does not re-run them when the base moves — the run above was triggered by a push against `develop`. Locally:

```
[extract-api] out of date: knowledge/api/authoring.json, … (12 files)
  regenerate: node knowledge/tools/api-surface/extract-api.mjs --from-reactor
```

Worth noting because it is the first time the gate caught something ordinary rather than something I broke on purpose. A base change produces no source diff and looks harmless; without the gate the pack would have merged claiming a version the tree does not have.

**Only the version stamps moved** — `verifiedAgainst`, `targetVersion`, `generatedFrom`. The surface is unchanged at 227 / 161 / 69 / 9 / 5 types with 163 exclusions, which is what a pom-only base commit should produce and is itself a check on the extractor: a version bump that shifted an API line would mean something reads the version where it should read the classes.

Rebase rather than a merge commit because this repository squash-merges — zero merge commits in the last 30 on `develop`, and #625–#631 each landed as a single-parent commit. The seven commits collapse to one at merge either way, so preserving their SHAs buys nothing and a merge commit would only add noise to the squash diff.

**Verification on the new base:** reactor slice `clean verify` → BUILD SUCCESS, **2098 tests across 375 classes, 0 failures, 0 errors**. Knowledge gates: `classifier.test` 23, `claims.test` 18, `extract-api --check` current, `check-claims --check` current, `check-routes` 3 routes, `build-bundle --verify` answers standalone.

